Issue #23, opened 2015-12-22
Laravel 5.2 config('auth.model') is gone


SERVANT14:
I noticed this after upgrading. OwenIt\Auditing\Log model has a relation that expects this to exist.

Here's the change for 5.2:
https://github.com/laravel/laravel/commit/8414d45cdc351e495016bc19ba2821a5d76c494e#diff-3fc9d8e570780c4eb337c6f860b1e8e0L22

It looks like it moved here: https://github.com/laravel/laravel/blob/develop/config/auth.php#L70

Here's the relationship in the Log model:
https://github.com/owen-it/laravel-auditing/blob/master/src/Log.php#L52

I'm not sure the proper way to get at this now, just thought I'd bring it up.

anteriovieira:
Thank you @SERVANT14 , I'll check.

anteriovieira:
Added the configuration file auditing, for now we solve our problems, but we have to mention in the documentation as well. This can be a good alternative, since in the future we may need more settings.

Fixed in PR https://github.com/owen-it/laravel-auditing/pull/24.

PR #24, opened 2015-12-23
Added configuration file

anteriovieira:
This PR allows the use of the configuration file for auditing.

````php
//Example:

return [
    'model' => App\User::class,
];

```


anteriovieira:
Related to https://github.com/owen-it/laravel-auditing/issues/23

PR #26, opened 2016-01-26
Some suggestions on first part of readme

SERVANT14:
Put together some suggestions 'cuz I said I would :). Let me know what you think. We can add the differentiation between revisionable in the future as well. I didn't have any good ideas on that to suggest yet.


anteriovieira:
Hi @SERVANT14 

I have nothing to say, they are all good suggestions. :)

Let's add the differentiation between revisionable in the future.

Thank you very much.

PR #46, opened 2016-05-22
Add fields for saving user's IP and current request route when logging.

achillesp:
I have added a new migration for the extra fields and updated AuditingTrait to save them to log.


anteriovieira:
Thank you

PR #86, opened 2016-08-15
Allows custom names for owner_type.

marcoskubis:
This update allows custom names for owner_type field by getting the morph class alias set by "Relation::morphMap()".


anteriovieira:
Thanks @marcoskubis

marcoskubis:
This causes a conflict with the custom messages. I will send a patch with the fix.

Issue #106, opened 2016-10-02
[Proposal] Ability to log that a field value has changed, but not store the actual value (e.g. password)


kitbs:
I would like to record an audit log when the User model has been updated if the password has been changed. However, I do not want to store the password/hash in my audit trail table.

If I were to use `$dontKeepAuditOf = ['password']` on my User model, then a change of password field would not be included in the audit log. And if the password was the only value changed, then no audit log would be created at all.

It would be nice if the auditing function looked at the `$hidden = ['password']` on my model, then stored `{"password": null}` in the old and new columns instead of the actual value. Or some other way of customising "hidden" attributes in the audit log.

What do you think?

anteriovieira:
@kitbs , I think a great idea. You like to send a pull request?

kitbs:
Sorry, my automatic indenting is slightly different to yours and was causing lots of unnecessary diffs. The final commit and PR just shows the changed lines.

I decided to make the behaviour respect `$hidden` by default (i.e. change the current behaviour) but it can be switched off by setting `protected $auditRespectsHidden = false;` on the model.

If you prefer that the current behaviour stays as the default, I could reverse the logic so that the new behaviour only works if `$auditRespectsHidden` is `true`. Up to you.

Thanks for the great package and the chance to contribute.

anteriovieira:
I think we should audit all values. If you wish to omit the hidden values, this should be explicit in the model.

PR #107, opened 2016-10-04
[3.0] Audited values will respect $hidden/$visible

kitbs:
Response to https://github.com/owen-it/laravel-auditing/issues/106
Behaviour can be ignored by setting protected `$auditRespectsHidden = false;` on the model.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -199,8 +199,8 @@ trait Auditable
         // Auditable data
         return [
             'id'             => (string) Uuid::uuid4(),
-            'old'            => $this->oldData,
-            'new'            => $this->newData,
+            'old'            => $this->cleanHiddenAuditAttributes($this->oldData),
+            'new'            => $this->cleanHiddenAuditAttributes($this->newData),
             'type'           => $this->typeAuditing,
             'auditable_id'   => $this->getKey(),
             'auditable_type' => $this->getMorphClass(),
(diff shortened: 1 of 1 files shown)


anteriovieira:
@kitbs , Perfect!

I would be grateful if you could update the documentation.

kitbs:
Hi @anteriovieira 

When I started on the documentation, I remembered that I ended up committing the behaviour that defaults to false, and will store the actual values of hidden fields unless specified. So what I said in https://github.com/owen-it/laravel-auditing/issues/106#issuecomment-251522020 is actually the opposite way around. :slightly_frowning_face: 

This will default to `false` if `$auditRespectsHidden` is not set on the model:

````php
public function isAuditRespectsHidden()
{
    return isset($model->auditRespectsHidden) && $model->auditRespectsHidden;
}
````

But whis will default to `true` if `$auditRespectsHidden` is not set on the model:

````php
public function isAuditRespectsHidden()
{
    return !isset($model->auditRespectsHidden) || $model->auditRespectsHidden;
}
````

Before I commit any documentation, would you prefer to change the default behaviour (will need a new PR) or leave it as is in this PR?

Sorry for the confusion.

anteriovieira:
@kitbs, no need to apologize.

I think we should audit all values. If you wish to omit the hidden values, this should be explicit in the model.

PR #142, opened 2016-10-31
[3.2] Added auditing many to many relationships for eloquent models

paulkned:
This pull request basically adds auditing for many to many relationships on Eloquent models. Since Eloquent doesn't support event hooks on many to many relationships out of the box I "borrowed" some code from a pull request which was denied on laravel core and made it suitable for this package. If these events are ever added (the reason for denying the pull request didn't make it seem that they ever would) this will need to be refactored. But for now it works like a charm.


anteriovieira:
Hi @paulkned , please send to master.

Issue #143, opened 2016-11-03
Custom audit message translations


quetzyg:
Hi there,

First of all, great package!

Now to business. Currently, this package doesn't have support for translated messages.
Also, the way the custom message and custom field are set seems messy, since we're setting them inside the model.

A better approach would be to move that into the `resources/lang/{your_language}/` path and have it set in the model like:
```php
// Booking.php model
public static $auditCustomMessage = 'booking.audit.message';

public static $auditCustomFields = 'booking.audit.fields';
```

Then, we would have a `resources/lang/en/booking.php` with:
```php
<?php

return [
    'audit' => [
        'message' => '{user.name} {type} this Booking {elapsed_time}',
        'fields'  => [
            'date'       => 'The date was defined as "{new.date}"',
            'seats'      => 'The seats were defined as "{new.seats}"',
            'ip_address' => 'Registered from the address {ip_address}',
        ],
    ],
];
```
In the `CustomAuditMessage` trait, the `getCustomMessage()` method would return the translated string:
```php
    public function getCustomMessage($class)
    {
        if (!isset($class::$auditCustomMessage)) {
            return trans('audit.custom.message');
        }

        return trans($class::$auditCustomMessage);
    }
```
and the `getCustomFields()` method would behave the same way:
```php
    public function getCustomFields($class)
    {
        if (!isset($class::$auditCustomFields)) {
            return [];
        }

        return trans($class::$auditCustomFields);
    }
```

Thoughts?

anteriovieira:
Hi @quetzyg , great idea. Feel free to send us a pull request.

quetzyg:
@anteriovieira, to which branch should I PR to? `master` or `3.1`?

anteriovieira:
@quetzyg send to master.

PR #144, opened 2016-11-04
feature/message-translations

quetzyg:
Hello!

This PR adds initial message translation support, as suggested in issue #143.

The changes are very simple and for the time being it should be fine, since the BC breaks are minimal.

I do however, suggest leveraging on the Symfony [translation component](http://symfony.com/doc/current/components/translation.html), which Laravel uses btw, to simplify the `CustomAuditMessage` trait.

But that can be done in another PR, since the syntax would have to change slightly: `{foo}` vs `:foo`.

**PS:** the documentation should also be updated to reflect changes, if this gets merged.

### DIFF
--- a/lang/en/auditing.php
+++ b/lang/en/auditing.php
@@ -1,0 +1,20 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Auditing Language Lines
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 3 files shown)


anteriovieira:
Hi @quetzyg , You can update the documentation?

quetzyg:
Sure, bear with me just a bit.

anteriovieira:
Thank you, keep up the good work.

PR #148, opened 2016-11-11
Removed User model from configuration

stfnndrsn:
Changed User model lookup from local configuration to use auth.providers.users.model provided from the Laravel framework.

Less configuration to do for the package - easier integration.


anteriovieira:
Thanks @stfnndrsn , can you update the documentation too?

stfnndrsn:
Done. https://github.com/owen-it/laravel-auditing-doc/pull/13

PR #176, opened 2016-12-23
Change insert method to create for enable eloquent events

marcoskubis:


anteriovieira:
@marcoskubis , please, the improvements should be sent to the master branch, lower branches just for bug fix.

Issue #184, opened 2017-01-20
Default audits to be returned ordered most recently first


alexdover:
Me again!

I am returning the audits for an object and think it would be worth ordering them by the created date.

I think it would be as simple as this:

    public function audits()
    {
        return $this->morphMany(Auditing::class, 'auditable')->orderBy('created_at', 'desc');
    }

in the DatabaseAudits trait

I'd be happy to put this in a pull request if it helps.

quetzyg:
Hi @alexdover,

Thanks for the suggestion, but I think we should leave it to the user to decide by which column and in which order the results should be returned.

You can easily achieve what you want with:
```php
        return $someModel->audits()
            ->orderBy('created_at', 'DESC')
            ->get();
```

@anteriovieira, what are your thoughts?

anteriovieira:
@alexdover  , feel free to send us a pull request.

@quetzyg seems to me interesting behavior.

quetzyg:
This feature will be present in the next version.

PR #188, opened 2017-01-29
feature/minor-updates

quetzyg:
In this pull request:
- Remove dependency on the `ramsey/uuid` package
- Return the `Audit` sorted by `created_at` in descending order
- Updated version requirements (support Illuminate libraries from 5.1 onwards)
- Added license file + headers

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -81,7 +81,8 @@ trait Auditable
      */
     public function audits()
     {
-        return $this->morphMany(AuditModel::class, 'auditable');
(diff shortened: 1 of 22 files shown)


src/Auditable.php
```diff
@@ -81,7 +92,8 @@ public static function bootAuditable()
      */
     public function audits()
     {
-        return $this->morphMany(AuditModel::class, 'auditable');
+        return $this->morphMany(AuditModel::class, 'auditable')
```
quetzyg: Fixes #184

PR #193, opened 2017-02-07
Interrogate database for counting rows for deletion only if model has historyLimit attribute set.

sbogx:
If the historyLimit attribute is not set on the model then an unused database query will run and this can impact performance throughout the application.


anteriovieira:
Fix the errors and resubmit the request.

Thanks!

PR #210, opened 2017-03-16
	Interrogate database for counting rows for deletion only if model has historyLimit attribute set

sbogx:
- added travis config from master and changed the versions for supported PHP to those defined on branch 3.1 (https://github.com/owen-it/laravel-auditing/blob/3.1/.travis.yml)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-    - 5.5
     - 5.6
     - 7.0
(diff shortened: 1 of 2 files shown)


anteriovieira:
Thanks!

sbogx:
Thank you too. I saw that the same implementation is in other tags. Would it helpful if I update them and create new pull requests?

PR #220, opened 2017-03-24
Fix migration stub to use configurations for database driver

the-great-tao:
This allows us to change database driver configuration before calling "php artisan migrate".

### DIFF
--- a/src/Console/stubs/audits.stub
+++ b/src/Console/stubs/audits.stub
@@ -25,17 +25,18 @@ class CreateAuditsTable extends Migration
      */
     public function up()
     {
-        Schema::create('audits', function (Blueprint $table) {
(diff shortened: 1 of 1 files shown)


anteriovieira:
Ping @quetzyg

quetzyg:
Nice spotted @the-great-tao,

I would just ask you to update your pull request to include the same logic in the `down()` method.

```php
    public function down()
    {
        Schema::connection(config('audit.drivers.database.connection'))
            ->drop(config('audit.drivers.database.table'));
    }
```

the-great-tao:
Done, please check. Thank you!

anteriovieira:
@the-great-tao  thanks for this PR.

PR #222, opened 2017-04-11
Added withTrashed to auditable relation and avoid logging item as updated when restoring

gmedeiros:
When retreiving a softdeleted item, I was getting an error, so I added withTrashed when at the auditable() relation. This is because I am auditing a model and some related models using many to many relationships (which required a few changes and not using the sync() method.

Also, when restoring a softdeleted item, an updated audit was being created. I added some logic to check if only the updated_at and deleted_at are changing (with deleted_at being not null) before saving an updated audit.

### DIFF
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -41,6 +41,7 @@ class AuditableObserver
     public function updated(AuditableContract $model)
     {
         if (
+            method_exists($model, 'getDeletedAtColumn') &&
             $model->isDirty($model->getDeletedAtColumn()) &&
             !$model->deleted_at &&
             count($model->getDirty()) == 2 //deleted_at and updated_at
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -40,7 +40,16 @@ class AuditableObserver
      */
     public function updated(AuditableContract $model)
     {
-        Auditor::execute($model->setAuditEvent('updated'));
+        if (
+            $model->isDirty($model->getDeletedAtColumn()) &&
+            !$model->deleted_at &&
+            count($model->getDirty()) == 2 //deleted_at and updated_at
+        ){
+            // only restoring softdeleted item, no values changed other than deleted_at an updated_at
+        }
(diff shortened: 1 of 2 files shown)


src/AuditableObserver.php
```diff
@@ -40,7 +40,16 @@ public function created(AuditableContract $model)
      */
     public function updated(AuditableContract $model)
     {
-        Auditor::execute($model->setAuditEvent('updated'));
+        if (
+            method_exists($model, 'getDeletedAtColumn') &&
+            $model->isDirty($model->getDeletedAtColumn()) &&
+            !$model->deleted_at &&
+            count($model->getDirty()) == 2 //deleted_at and updated_at
```
quetzyg: While this may work, I don't like how the code is written. Specially the empty `if` part.

src/Models/Audit.php
```diff
@@ -93,7 +93,7 @@ public function getTable()
      */
     public function auditable()
     {
-        return $this->morphTo();
+        return $this->morphTo()->withTrashed();
```
anteriovieira: You can do this in your own code `$audit->auditable()->->withTrashed()`  😉

anteriovieira:
@gmedeiros, and if you ignore these properties, could it help?

```php
<?php

namespace App;

use OwenIt\Auditing\Auditable;
use OwenIt\Auditing\Contracts\Auditable as AuditableContract;

class Post extends Model implements AuditableContract;
{
    use Auditable;

    /**
     * Attributes to exclude from the Audit.
     *
     * @var array
     */
    protected $auditExclude = [
        'updated_at', 'deleted_at'
    ];
}
```

src/Models/Audit.php
gmedeiros: Sounds like a good idea, thanks!

quetzyg:
Thanks for the pull request @gmedeiros.

A cleaner implementation has been done in #233.
While your code didn't get merged, your PR was still helpful as a bug report.

This PR is now closed.

PR #238, opened 2017-05-08
Change URL type string to text

Cannonb4ll:
I had the problem when using socialite facebook the route url was to long for the database. I changed the type to text so the URL could fit.

Error message:

`SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'url' at row 1`


quetzyg:
Hi @Cannonb4ll,

While the default migration has the `url` column type set to `VARCHAR(255)`, it doesn't mean you can't change it.

Like stated in the [troubleshooting](http://www.laravel-auditing.com/docs/4.0/troubleshooting) section of the documentation, you're free to tweak it as you see fit.

For that reason, I'm going to reject this PR.

Still, thanks for the interest!

Issue #206, opened 2017-03-14
Audit presentation


hafizhassan:
i tried to display the record.

**this is my model.**
namespace App;
use OwenIt\Auditing\Models\Audit as AuditModel;
class Audit extends AuditModel
{  }

**this is my controller** 
public function index()
{   
   $audits = Audit::all();
   return view('audit.index', compact('audits'));
 }

**this is the blade**
<ul>
    @forelse ($audits as $audit)
    <li>
        @lang('post.updated.metadata', $audit->getMetadata())

        @foreach ($audit->getModified() as $attribute => $modified)
        <ul>
            <li>@lang('post.'.$audit->event.'.modified.'.$attribute, $modified)</li>
        </ul>
        @endforeach
    </li>
    @empty
    <p>@lang('post.unavailable_audits')</p>
    @endforelse
</ul>

**when run it show** 
FatalErrorException in Audit.php line 167:
Call to a member function hasGetMutator() on null

what did i miss?
thank for the attention. thank you.

quetzyg:
Have a look at the documentation, more specifically the [Getting Audits](http://www.laravel-auditing.com/docs/4.0/getting-audits) section.

quetzyg:
Basically, you should get `Audit` records through a Model:

```php
$audits = $someModel->audits();
```
rather than calling `all()` from the `Audit` model itself.

hafizhassan:
thank. ive miss something.

mikebronner:
I was running into the same issue. Is there a way to get all audit entries? I have an admin page that show all updates, regardless of model.

amendozaaguiar:
> i tried to display the record.
> 
> **this is my model.**
> namespace App;
> use OwenIt\Auditing\Models\Audit as AuditModel;
> class Audit extends AuditModel
> { }
> 
> **this is my controller**
> public function index()
> {
> $audits = Audit::all();
> return view('audit.index', compact('audits'));
> }
> 
> **this is the blade**
> 
> @forelse ($audits as $audit)
> * @lang('post.updated.metadata', $audit->getMetadata())
>   ```
>       @foreach ($audit->getModified() as $attribute => $modified)
>       <ul>
>           <li>@lang('post.'.$audit->event.'.modified.'.$attribute, $modified)</li>
>       </ul>
>       @endforeach
>   </li>
>   @empty
>   <p>@lang('post.unavailable_audits')</p>
>   @endforelse
>   ```
> 
> **when run it show**
[...]

PR #244, opened 2017-05-16
Fix iterating over all audits, if audible model is not resolved

mikebronner:
Possible fix for #206, not tested for side-effects, but seems to run OK with local dataset.

### DIFF
--- a/src/Models/Audit.php
+++ b/src/Models/Audit.php
@@ -162,7 +162,7 @@ class Audit extends Model
         $value = $this->data[$key];
 
         // Apply a mutator or a cast the Auditable model may have defined
-        if (starts_with($key, ['new_', 'old_'])) {
+        if ($this->auditable && starts_with($key, ['new_', 'old_'])) {
             $originalKey = substr($key, 4);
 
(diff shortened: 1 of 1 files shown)


anteriovieira:
Ping @quetzyg

quetzyg:
Hi @mikebronner,

While I understand the problem (`Auditable` morphTo relation not being loaded), I'm unable to reproduce it.

I currently have the `audits` table with quite a few entries from several model types and I'm doing:
```php
    foreach (\OwenIt\Auditing\Models\Audit::all() as $audit) {
        dump($audit->getModified());
    }
```
I'm able to get everything as expected, never getting the issue described in #206.

Can you show me how you're running into it?

quetzyg:
Actually, I just read @truonggiangdao's comment about deletion, which makes total sense now.

mikebronner:
Thanks for merging! Yea, I'm pretty sure the issue was caused by deleting database entries. Sorry I wasn't more explicit in my description :). (Was focused on the symptom, instead of also researching the cause.)

quetzyg:
Not a problem mate, thank you both for reporting!

Issue #58, opened 2016-06-01
Not logging Affected Relations?


bbredewold:
Is it correct that any updated relations are not being logged in the database?
I'm using eloquent's sync() function a lot, for updating many-to-many relations...

Any ideas how to do that?

camih:
The auditing currently uses events for, but in case of pivots, there are no events being triggered. There are many issues reported regarding this. See for example here: https://github.com/laravel/framework/issues/8452 . 
I also need this and currently did not find a very nice way to handle it.

robclancy:
Can you not just override those methods like sync, attach and detach? Or even better would be a chain on the relationship in some way, will explain more if I come up with a way of doing this. Trying out this package and I think we will need it to work for this and we can work around it easy enough, if a workaround is good enough we will make a PR.

Jono20201:
Been having this issue and have managed a work around, but don't really like having to override Model methods and copy/paste code from the Laravel source when over-riding. Managed to extend BelongsToMany class to over-ride the attach/deattach method however.

Tempted to try and make a PR as well. I really think Laravel needs to trigger events for this.

anteriovieira:
Sorry for the delay.

Hello @robclancy , let's think of something. I'll keep this open to give continuity to this improvement.

fly-studio:
yes, 'sync attach detach' donot log, 
4.x support it?

quetzyg:
@fly-studio, as soon as version 4.0 is out the door, we'll try to work on many-to-many audit support for 4.1.

rorc:
When will release 4 version? 😃

quetzyg:
@rorc, when it's done (i.e. when I finish adding tests)

Meanwhile, you already can use `"owen-it/laravel-auditing": "4.0.x-dev"` in your `composer.json`.
The `v4` documentation is available [here](http://www.laravel-auditing.com/docs/master).

rorc:
@quetzyg Ok, thanks 😄

jschlies:
I strongly vote for this feature. I'm currently working on a custom auditor that would record (to a seperate table) the foreign key relationships relations to  a separate table, then I extend $Model->getAuditArr() to pull it all together the way we want. Lots of DB hits and in general, not a perfect solution. Will switch off this when this feature is implemented.

rorc:
How progress here? 😀

od3n:
+1 still waiting for the best solution

daniel-gomes-sociomantic:
[...]

PR #257, opened 2017-06-25
Related relations

jschlies:
This pull request is my initial attempt to audit the relationships of a particular model. Please comment but I am unwilling to push this intil I deal with the 3 Todo's.

Be warned: This can have negative impact on DB performance. Be very cautious to not use this feature on too many models (and those models relationships). This PR offers controls on both.

Todo: 
Create table related_relations and use this rather that audits:related_relation_json
Integrate this into $model->audits()
Better unit tests

Todo (far future)
Make this Event/Queue friendly to try to deal w/ the DB load this this can generate.

To use:

Run the migration adding related_relation_json to audits table

All models for which you want to track related_relations need: 

1. A public property of $model->auditIncludeRelated set to true
1. These public methods: $model->getBelongsToArr(), $model->HasManyArr,$model->GetBelongsToManyArr(), $model->getHasOneArr(). These should return an array of strings that indicate the names of the Eloquent properties (relations) you wish to track.
1. If you are auditing the related models of $a (say $b and $c), you minimally need to audit $b and $c.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -200,24 +200,101 @@ trait Auditable
         $this->updateAuditExclusions();
 
         $old = [];
-        $new = [];
+        $new              = [];
+        $related_relations_arr = $this->generateRelatedRelations();
 
         $this->{$method}($old, $new);
 
-        return $this->transformAudit([
-            'old_values'     => $old,
-            'new_values'     => $new,
-            'event'          => $this->auditEvent,
-            'auditable_id'   => $this->getKey(),
-            'auditable_type' => $this->getMorphClass(),
-            'user_id'        => $this->resolveUserId(),
-            'url'            => $this->resolveUrl(),
-            'ip_address'     => $this->resolveIpAddress(),
-            'user_agent'     => $this->resolveUserAgent(),
-            'created_at'     => $this->freshTimestamp(),
-        ]);
+        return $this->transformAudit(
+            [
+                'old_values'             => $old,
+                'new_values'             => $new,
+                'event'                  => $this->auditEvent,
+                'auditable_id'           => $this->getKey(),
+                'auditable_type'         => $this->getMorphClass(),
+                'user_id'                => $this->resolveUserId(),
+                'url'                    => $this->resolveUrl(),
+                'ip_address'             => $this->resolveIpAddress(),
+                'user_agent'             => $this->resolveUserAgent(),
+                'related_relations_json' => json_encode($related_relations_arr),
+                'created_at'             => $this->freshTimestamp(),
+            ]
+        );
     }
 
+    public function generateRelatedRelations()
+    {
+        $related_relations_arr        = [];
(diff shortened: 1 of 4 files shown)


quetzyg:
Hi @jschlies,

Thanks for trying to come up with a solution to the many-to-many relation auditing.

On your TODO, you state that there should be a `related_relations` table, instead of a new `audits` column (`related_relation_json` in this case), why so?

Also, the logic you're proposing is using relation types (`BelongsTo`, and so on) instead of events, why?

And how do you deal with a `sync()` call that would remove/add lots of relations? Would you create an audit record for each, or do you create a single one with all the changes?

jschlies:
@quetzyg 

> On your TODO, you state that there should be a related_relations table, instead of a new audits column (related_relation_json in this case), why so?

It would be extremely hard to extract the data needed to run $somethingRelatedToodel->audits() for not only the object in question but it's relations. As I stated above, this fix runs the risk of overloading the DB w/ insert requests. Perhaps there is a way to bundle these inserts or get them offline somehow. 

> Also, the logic you're proposing is using relation types (BelongsTo, and so on) instead of events, why?

This is our current implementation and it serves us well. I will explore using events as you indicate. Can you point me in a direction on this?

 > And how do you deal with a sync() call that would remove/add lots of relations? Would you create an audit record for each, or do you create a single one with all the changes?

I don't but if I did, (and I will research doing so), each relationship would get it's own related_relation record.  As I stated above, this fix runs the risk of overloading the DB w/ insert requests. Perhaps there is a way to bundle these inserts or get them offline somehow. 

-=-=-=-=-=-=-
Thanks for the input and will explore the issues you point out. Keep the comments coming

jschlies:
See attached for our current implementation of $this->getAuditArr(). Ignore the sorting stuff.

quetzyg:
>This is our current implementation and it serves us well. I will explore using events as you indicate. Can you point me in a direction on this?

In regards to the usage of events, I [commented](https://github.com/owen-it/laravel-auditing/issues/58#issuecomment-306159633) in the open issue #58. The idea would be to fire an event from a repository and that would kickstart the audit process.

quetzyg:
You totally lost me with the `AuditableTrait.php.txt`

jschlies:
I think that a great idea (events and potentially queuing) and will implement but I think that a separate topic to be taken up later.

jschlies:
AuditableTrait.php.txt contains our implementation of getAuditArr().
Can you elaborate?

quetzyg:
The code, is confusing.

jschlies:
I'll spend some time commenting. 
Keep in mind:

- The table I propose, related_relations, is called audit_relations in our implementation. 
- $generations keeps our use of recursion under control.
- If you think of  a beter nomenclature than 'source' and 'related', please let me know.
- In our implementation, we only really care about belongsTo.
- ignore the sorting stuff

quetzyg:
Closing it, since we're currently working on version 4.1.

PR #261, opened 2017-07-03
Updates and fixes

quetzyg:
Work done in this pull request:
- Added support for Laravel 5.5 auto-discovery
- Fix `Audit` **resolveData()** method (use `User` foreign key value)


quetzyg:
In order to keep things simple and clean, the `Config::get()` calls in the migration were later removed. If you need to change something in the migration, go ahead. We even have a specific [section](http://www.laravel-auditing.com/docs/8.0/audit-migration) in the documentation with examples.

PR #276, opened 2017-07-17
Ensure $this->user is not null

Arkitecht:
I ran into an issue where if I modify an audited model without a user (ie. from the CLI) and then try and grab the audits it throws:
`PHP Fatal error:  Call to a member function attributesToArray() on null in [...]/vendor/owen-it/laravel-auditing/src/Models/Audit.php on line 128`

This modification ensures there is a user set before calling `$this->user->attributesToArray()` therefor avoiding the fatal error


quetzyg:
Thank's for this PR @Arkitecht.

I'll do some tests and try to reproduce the issue locally.

PR #289, opened 2017-08-12
Related relations

jschlies:
NOT READY FOR MERGE
I'm going to try to give this some of my time. I hope I have a 'merge-able' branch in a few months. This PR is my work so far. 
Until ready for merge, please make comments in https://github.com/owen-it/laravel-auditing/issues/58 
Going forward, I want to:

create a separate table (audit_related_relations) which will (on eloquent.synced*, eloquent.attached* and eloquent.detached*) record the One To One, One To Many and Many To Many. Has Many Through, Polymorphic Relations and Many To Many Polymorphic Relations are going to be phase II.
Develop a system/convention/config that limits which objects get audit_related_relations, do not get audit_related_relations or get audit_related_relations via a queued Job,  and for the object in question, which relationships get audit_related_relations, do not get audit_related_relations or get audit_related_relations via a queued Job. This limited is important since it is easy to see how this can explode into huge amounts of processing.
Develop a system/convention/config that limits the size/depth of $audit object.  This limited is important since it is easy to see how this can explode into huge amounts of processing and extremely large objects.


quetzyg:
I'm not a fan of using a second audit table. You also talk about a convention to limit what gets audited and what doesn't. I also don't like that idea. Read my [comment](https://github.com/owen-it/laravel-auditing/issues/58#issuecomment-306159633) about letting the developer decide which data should be audited.

PR #291, opened 2017-08-17
Change user_agent column to be text, not string

nickysemenza:
User Agents aren't necessarily fewer than 255 chars. I've been getting a few `Data too long for column 'user_agent'` exceptions on productions, and that causes a method to exit out after `->save()` is called on the modal 😢


quetzyg:
You just have to change the migration to suit your needs. The [troubleshooting](http://www.laravel-auditing.com/docs/4.1/troubleshooting) guide has quite a few suggestions about it.

PR #295, opened 2017-08-22
Remove static reference to OwenIt\Auditing\Models\Audit in Database driver

walmsles:
Database driver still constructs static OwenIt\Auditing\Models\Audit class instead of using the configured Model class in 'audit.implementation' section of config file.  

This is required so I can override the Audit model and store Audit trail in Mongo database store using my own model implementation.


quetzyg:
Good catch!

Issue #297, opened 2017-08-24
Eloquent MongoDB object changes not auditing


bsormagec:
I am using it with MongoDB Driver and its not working when mongodb object has been changed. So its not working recursive only checking first level of array.

quetzyg:
I don't want to open this can of worms, so for now we won't add support for this.

PR #298, opened 2017-08-24
fix recursive array issue

bsormagec:
Fixed if a field is an array then check it and save differency to audit table this is required when you use cast like object or array.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -132,7 +132,7 @@ trait Auditable
         }
     }
 
-    function recursiveUpdatedAttributes(array $value, $attribute, &$old, &$new)
+    public function recursiveUpdatedAttributes(array $value, $attribute, &$old, &$new)
     {
         foreach ($value as $attr => $val) {
             if ($this->isAttributeAuditable($attribute)) {
@@ -211,15 +211,15 @@ trait Auditable
         $foreignKey = Config::get('audit.user.foreign_key', 'user_id');
 
         return $this->transformAudit([
-            'old_values' => $old,
-            'new_values' => $new,
-            'event' => $this->auditEvent,
-            'auditable_id' => $this->getKey(),
-            'auditable_type' => $this->getMorphClass(),
-            $foreignKey => $this->resolveUserId(),
-            'url' => $this->resolveUrl(),
-            'ip_address' => $this->resolveIpAddress(),
-            'user_agent' => $this->resolveUserAgent(),
+            'old_values'        => $old,
+            'new_values'        => $new,
+            'event'             => $this->auditEvent,
+            'auditable_id'      => $this->getKey(),
+            'auditable_type'    => $this->getMorphClass(),
+            $foreignKey         => $this->resolveUserId(),
+            'url'               => $this->resolveUrl(),
+            'ip_address'        => $this->resolveIpAddress(),
+            'user_agent'        => $this->resolveUserAgent(),
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


quetzyg:
Can you provide an example of what this is trying to solve?

bsormagec:
I am using Mongodb and using object fields on my models and when i changed a value in object laravel-auditing can't detect that change its recording a nothing changed record. With that if we have an array type or object type cast in our Eloquent Model now detect changes. 

In User Model I have a field settings with type(cast) object when i changed something in that object auditing cant detect it. with that change it will detect now.

bsormagec:
fix #297

src/Auditable.php
```diff
@@ -131,9 +131,16 @@ protected function auditCreatedAttributes(array &$old, array &$new)
     protected function auditUpdatedAttributes(array &$old, array &$new)
     {
         foreach ($this->getDirty() as $attribute => $value) {
-            if ($this->isAttributeAuditable($attribute)) {
-                $old[$attribute] = array_get($this->original, $attribute);
-                $new[$attribute] = array_get($this->attributes, $attribute);
+            if (is_array($value)) {
+                foreach ($value as $attr => $val) {
+                    $old[$attribute][$attr] = array_get($this->original['settings'], $attr);
```
quetzyg: What is this `settings` index?

src/Auditable.php
```diff
@@ -131,9 +131,16 @@ protected function auditCreatedAttributes(array &$old, array &$new)
     protected function auditUpdatedAttributes(array &$old, array &$new)
```
quetzyg: What if there are more than 2 levels in the array?

bsormagec: i fixed my problem for now and maybe some other users, it will be fix in future by someone or i will try to fix it soon.

quetzyg: If we're gonna change this, I prefer to have it done for any (sane) level number, instead of being a quick fix for a user's use case.

bsormagec: I fixed it for unlimited level

bsormagec:
i fixed all issues

src/Auditable.php
```diff
@@ -91,16 +91,6 @@ protected function updateAuditExclusions()
                 $this->auditableExclusions[] = static::DELETED_AT;
             }
         }
-
```
quetzyg: I won't accept this removal.

src/Auditable.php
```diff
@@ -131,9 +121,23 @@ protected function auditCreatedAttributes(array &$old, array &$new)
...
+            } else {
+                if ($this->isAttributeAuditable($attribute)) {
+                    $old[$attribute] = array_get($this->original, $attribute);
+                    $new[$attribute] = array_get($this->attributes, $attribute);
+                }
+            }
+        }
+    }
+
+    public function recursiveUpdatedAttributes(array $value, $attribute, &$old, &$new)
```
quetzyg: Not a fan of this approach.

walmsles:
Given this package's focus is on auditing changes to Eloquent models and you require recursive auditing over array contents within your Mongo Model - would it be sensible to alter the way you setup your models and utilise embedded related models within your code instead (assuming you are using jessengers/mongodb package for laravel)?

I know this would allow you to keep an embedded structure but the models would be separate eloquent class instances and would likely (not sure myself yet as haven't tested this scnenario sepcifically) be simply audited as is by the default package with no change?

I think this would be a neater approach and more in line with clean laravel coding.

Just a thought

Issue #198, opened 2017-03-03
error when route is longer than 250 char


tunele:
There's an error when route is longer than 250 char, for example because of a many ulr vars.
I think that the route value might be truncated to a maximum of 250. Anyways, I think that an audit error should not affect the proper behaviour of the app. Or at least the plugin should provide a configuration option: break on audit error true or false.

SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'route' at row 1

quetzyg:
Hi @tunele,

If you work with huge URL's, that's something you'll have to account for. Just update the migration and increase the `VARCHAR` value or change the type to `TEXT`.

This issue has been already discussed [here](https://github.com/owen-it/laravel-auditing/issues/174) so I'll mark this as closed.

Issue #174, opened 2016-12-15
Exclude querystring parameters from route column


marcroberts:
I'm using Socialite to login via Facebook and create a remember_token for the user, this creates an audit entry for the change of `remember_token` however the long callback URL is causing this error:

```
SQLSTATE[22001]: String data, right truncated: 7 ERROR: value too long for type character varying(255) (SQL: insert into "audits" ("id", "old", "new", "type", "auditable_id", "auditable_type", "user_id", "route", "ip_address", "created_at") values (aaf7d681-6232-4564-b136-abcbfe737266, {"remember_token":""}, {"remember_token":"MV0r0V0sUNInK7GltbXnudI5lxLYvRfDUT2zJgLZZcVYAxRR17us6R89awVJ"}, updated, 22995, App\User, 22995, http://xxxxxx.dev/auth/facebook/callback127.0.0.1code=AQ*SNIP**xx&state=xx**SNIP**xx, 2016-12-14 17:51:49, ?))
```

It looks like there are a couple of issues here:

1) The URL being used is actually too long (I've trimmed the long random strings out of it above), because of the querystring parameters. Should/could these be ignored?

2) The question mark in the URL appears to be being interpreted as a statement parameter with the IP address being replaced in to it. If the querystring is going to remain (and the data field increased in size to cope with it) what can we do about this?

Versions:
laravel-auditing: 3.1.10
laravel/framework: 5.3.26
PHP: 7.0.13

anteriovieira:
Hi @marcroberts ,

By default, [columns of type string](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Database/Schema/Blueprint.php#L464) is defined with 255 character 

I suggest you [change the column type route to text](https://github.com/laravel/framework/blob/5.3/src/Illuminate/Database/Schema/Blueprint.php#L475).

@quetzyg Let's fix this for next version.

quetzyg:
Hello @marcroberts,
1) That's an easy fix. You just have to change the column type from `VARCHAR(255)` to `TEXT`. We'll make `TEXT` as the default in the next version.

2) I'm not sure I understand what's happening. Can you be a bit more clear, please?

marcroberts:
1. I have done that for now, but there's a lot of useless querystring data being added that isn't useful (in the instance above, the code and state for the facebook callback are huge and pointless to store). Is there any way to exclude/ignore the querystring parameters from being inserted in this column?

[...]

PR #302, opened 2017-09-13
Fix url field too short bug

lasseal:
Change field type to 'text' to increase allowed length.

Fails with 'string' type when using Laravel Socialite
to login to Facebook which returns a string which
is > 400 characters. Changing to 'text' field type
allows more than the 'string' type 255 charachters.

screen shot 2017-09-13 at 07 04 05


quetzyg:
This has been discussed in #174, #198 and #238 (at least). There's also reference to this type of thing in the [troubleshooting](http://www.laravel-auditing.com/docs/4.1/troubleshooting) section of the documentation.

PR #321, opened 2017-09-30
related_relations

jschlies:
Add ability to collect and retrieve audits on related objects. 

Example: Say you need to maintain an accessLists to define which **Users** have access to which **Resources**. You could create the following database tables/models:  AccessLists, Users and Resources,with 'connector' tables/models AccessListUsers and AccessListResources. You could, with the current functionality, audit all 5 tables and somehow cobble together a picture if the full history of a particular AccessList but that picture would have issues. Say a user is added, then deleted and then re-added. Using only current functionality, it would be very hard to create the history of that AccessList that includes the Users first membership.

Related Relations seeks to solve this.

Indicate the model(s) (relating model) that you want to retain related relations for. For each, indicate the 'propertys' or 'methods' of that relating model that returns returns related objects. 'Yields' is the type of object returned. This results in $AuditObj's with a 'event' of 'related' and (with a call to $AuditObj->getRelatingAudit), the audit object of that is related.

To implement the above example using Related Relations to record activity :
Add the following to the AccessListUsers and AccessListResources  models.

class AccessListUsers extends Model implements AuditableContract, UserResolver
{
    use AuditableTrait;

    protected $auditRelatedProperties = [  'accessList'];
}
class AccessListResources extends Model implements AuditableContract, UserResolver
{
    use AuditableTrait;

    protected $auditRelatedProperties = [  'accessList'];
}

With this, when AccessListUsers or  AccessListResources are added, deleted or updated, an additional 'related' audit of the AccessLists model is taken and tied to the AccessListUsers or AccessListResource audit via relatuion_id.

Thus after adding an AccessListUsers, AccessLists->audits() will return an array of audit objects, One of which will have an 'event' value of related. The 'related' thing is accessable from that audit via Audit->getRelated().

Note that 'relation' here means more that Eloquent relation. $auditRelatedProperties needs to be a list of methods or properties that return an Eloquent object (or collection of objects) that have the Auditable trait and return an  obg. the  methods or properties. 
 
BE CAREFUL!!!! Be thoughtful of which relations you try to audit. This can create HUGE db traffic and should be used in a limited manner. If stress testing for any implementation of this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,8 +42,7 @@
         "php": ">=5.5.9",
         "illuminate/console": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
         "illuminate/database": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
-        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
-        "webpatser/laravel-uuid": "2.0.*"
+        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0",
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,8 @@
         "php": ">=5.5.9",
         "illuminate/console": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
         "illuminate/database": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
-        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*"
+        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
+        "webpatser/laravel-uuid": "2.0.*"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.0",
--- a/config/audit.php
+++ b/config/audit.php
@@ -78,4 +78,49 @@ return [
     */
 
     'console' => false,
+
+    /*
+     | Indicate the models (relating model) that you want to retain related relations for. For each,
+     | indicate the 'property' of that relating model that returns returns related objects. 'Yields' is the
+     | type of object returned.
+     |
+     | This results in $AuditObj's with a type of 'related' and a call to $AuditObj->getRelatingAudit.
+     |
+     */
+    'relation_hierarchy' => [
+        //\App\Models\Property::class => [
+        //    [
+        //        'property' => 'accessListProperties',
+        //        'yields' => \App\Models\AccessListProperty::class,
+        //    ],
+        //    [
+        //        'property' => 'client',
+        //        'yields' => \App\Models\Client::class,
+        //    ]
+        //],
+        //\App\Models\User::class => [
+        //    [
+        //        'property' => 'accessListUsers',
+        //        'yields' => \App\Models\AccessListUser::class,
+        //    ]
+        //],
+        //\App\Models\AccessList::class => [
+        //    [
+        //        'property' => 'accessListUsers',
+        //        'yields' => \App\Models\AccessListUser::class,
+        //    ],
+        //    [
+        //        'property' => 'accessListProperties',
+        //        'yields' => \App\Models\AccessListProperty::class,
+        //        'relator'=> 'property'
+        //    ],
+        //],
+        //\App\\Models\PropertyGroup::class => [
+        //    [
+        //        'property' => 'propertyGroupProperty',
+        //        'yields' => \App\Models\PropertyGroupProperty::class,
+        //        'relator'=> 'propertyGroup'
+        //    ],
+        //]
+    ]
 ];
--- a/config/audit.php
+++ b/config/audit.php
@@ -87,6 +87,8 @@ return [
      | This results in $AuditObj's with a 'event' of 'related' and (with a call to $AuditObj->getRelatingAudit), the
      | audit object of that is related.
      |
+     | BE CAREFUL. TEST CAREFULLY. This can have a very negative effect on performance.
+     |
      */
     'relation_hierarchy' => [
         //\App\Models\Property::class => [
@@ -112,5 +114,5 @@ return [
         //        'relator'=> 'property'
         //    ],
         //]
-    ]
+    ],
 ];
--- a/config/audit.php
+++ b/config/audit.php
@@ -84,7 +84,8 @@ return [
      | indicate the 'property' of that relating model that returns returns related objects. 'Yields' is the
      | type of object returned.
      |
-     | This results in $AuditObj's with a type of 'related' and a call to $AuditObj->getRelatingAudit.
+     | This results in $AuditObj's with a 'event' of 'related' and (with a call to $AuditObj->getRelatingAudit), the
+     | audit object of that is related.
      |
      */
     'relation_hierarchy' => [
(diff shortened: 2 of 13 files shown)


src/Audit.php
```diff
@@ -179,4 +179,80 @@ public function getModified($json = false, $options = 0, $depth = 512)
 
         return $json ? json_encode($modified, $options, $depth) : $modified;
     }
+
+    /**
+     * @return return all audits that were related to $this
+     */
+    public function getRelatedAudits()
```
quetzyg: Christ, you managed to use `camelCase`, `snake_case` and `PascalCase` in the same file!

src/Audit.php
```diff
@@ -179,4 +179,80 @@ public function getModified($json = false, $options = 0, $depth = 512)
...
+        }
+        /** @var \OwenIt\Auditing\Models\Audit $RelatingAuditObj */
+        $RelatingAuditObj = $audit_class::where('relation_id', '=', $this->relation_id)
+                                           ->where('event', '!=', 'related')->get()->first();
+        $RelatingAuditObj->setIsRelating(true);
+
+        return $RelatingAuditObj;
+    }
+
+    public $is_relating = false;
```
quetzyg: Attributes go on top!

src/Audit.php
```diff
@@ -179,4 +179,80 @@ public function getModified($json = false, $options = 0, $depth = 512)
...
+            'id'             => $this->id,
+            'ip_address'     => $this->ip_address,
+            'new_values'     => $this->new_values,
+            'old_values'     => $this->old_values,
+            'relation_id'    => $this->relation_id,
+            'updated_at'     => $this->updated_at,
+            'url'            => $this->url,
+            'user_agent'     => $this->user_agent,
+            'user_id'        => $this->user_id,
+           'related_audits'  => $this->event !== 'related' && !$this->isRelating() && $this->getRelatedAudits() ? $this->getRelatedAudits()->toArray() : null,
```
quetzyg: This takes too long to decipher.

src/Auditor.php
```diff
@@ -79,10 +81,49 @@ public function execute(AuditableContract $model)
             return;
         }
 
-        if ($audit = $driver->audit($model)) {
+        $uuid = null;
+        $list_of_properties = [];
+        foreach (config('audit.relation_hierarchy', []) as $class_type_to_audit => $local_list_of_methods) {
+            foreach ($local_list_of_methods as $local_method) {
+                if ($model instanceof $class_type_to_audit) {
```
quetzyg: What is this?

src/Auditor.php
```diff
@@ -79,10 +81,49 @@ public function execute(AuditableContract $model)
...
+                    $list_of_properties[] = $local_method['property'];
+                    $uuid = (string) Uuid::generate();
+                } elseif ($model instanceof $class_type_to_audit) {
```
quetzyg: You're doing a `if()` statement followed by an `elseif()` with the exact same thing!

composer.json
```diff
@@ -42,7 +42,8 @@
         "php": ">=5.5.9",
         "illuminate/console": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
         "illuminate/database": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
-        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*"
+        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
+        "webpatser/laravel-uuid": "2.0.*"
```
quetzyg: How is this better than https://github.com/ramsey/uuid ? And why is this needed?

config/audit.php
```diff
@@ -78,4 +78,41 @@
...
+     | indicate the 'property' of that relating model that returns returns related objects. 'Yields' is the
+     | type of object returned.
+     |
+     | This results in $AuditObj's with a 'event' of 'related' and (with a call to $AuditObj->getRelatingAudit), the
+     | audit object of that is related.
+     |
+     | BE CAREFUL. TEST CAREFULLY. This can have a very negative effect on performance.
+     |
+     */
+    'relation_hierarchy' => [
```
quetzyg: I'm not a fan of this.

src/Auditor.php
```diff
@@ -79,10 +81,49 @@ public function execute(AuditableContract $model)
...
+                    }
+                }
+            }
+        }
+
+        if ($audit = $driver->audit($model, $uuid, false)) {
             $driver->prune($model);
         }
 
+        if ($uuid) {
```
quetzyg: Nope.

src/Audit.php
jschlies: That was unnecessary. If there is a convention you want me to follow, please let me know and I will happily conform..

src/Audit.php
jschlies: OK, again if there is a list of conventions beyond continuous-integration/styleci/pr, please let me know and I will happily conform.

src/Auditor.php
jschlies: Please elaborate? I'm using is because I need to logically group the audit and 'related' audit records generated by a particular change to a model.

src/Audit.php
quetzyg: Like pretty much everybody else, we conform to PSR-2. And while not explicit, if you conform to the style of the existing code, you'll be able to figure that out.

src/Audit.php
quetzyg: Sorry, but this one is a no brainer. Again, look at the existing codebase.

src/Audit.php
jschlies: BTW, our in house convention is non-objects (ints, strings, bools) are snake case, methods in camelcase and objects in Pascal case.

src/Audit.php
jschlies: I'll break it up.

src/Auditor.php
jschlies: My bad. Thanks for the catch.

src/Audit.php
quetzyg: This is a different house, innit.

jschlies:
I will attempt to address your concerns in the coming days. If you want to see specific change or a specific technique, please be specific. I'll happily accept a PR, a patch file or other. 

I think the need/merit for this is covered in https://github.com/owen-it/laravel-auditing/issues/58. I will also create some doc to elaborate on the need and the approach.

jschlies:
I think this addresses your concerns. If I missed something or mis-interpreted something, please let me know. 

As for my general approach and the efficacy of this PR, I'll say that we have been using this approach for some time with success. I am not claiming that this is the absolute best solution nor that it does not have issues. For example, if not used judiciously, you can CRUSH your DB performance. Further, the 'event' based approaches suggested elsewhere have merits but I leave those improvements to others.

jschlies:
If this is waiting for some action on my part, please let me know.

quetzyg:
Sorry, I've been busy with work and other things. Let's postpone this, since I'm still not convinced about the approach with relating audits with other audits.

jschlies:
OK

quetzyg:
Like previously stated, I don't like the approach in this pull request.

jschlies:
OK, like I stated, this is working fine for us. Anywho, I'll watch this space for changes to address https://github.com/owen-it/laravel-auditing/issues/58 and I will keep this change on my github fork https://github.com/jschlies/laravel-auditing in case others need a solution for this in the mean time.

Thanks

PR #326, opened 2017-10-10
Added a filesystem driver

ymslavov:
This is being submitted as a preliminary request for comments. It is currently functional, both with local and remote disks, but I'm open to suggestions for improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-<a href="http://www.laravel-auditing.com" target="_blank"><img width="130" src="http://www.laravel-auditing.com/assets/img/laravel-auditing-logo.png"></a>
+<a href="http://www.laravel-auditing.com" target="_blank"><img width="130" src="http://www.laravel-auditing.com/assets/img/laravel-auditing-logo.png#v2"></a>
 </p>
 
 <p align="center">
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-<a href="https://www.laravel-auditing.com" target="_blank"><img width="130" src="http://www.laravel-auditing.com/assets/img/laravel-auditing-logo.png"></a>
+<a href="http://www.laravel-auditing.com" target="_blank"><img width="130" src="http://www.laravel-auditing.com/assets/img/laravel-auditing-logo.png"></a>
 </p>
 
 <p align="center">
--- a/README.md
+++ b/README.md
@@ -5,9 +5,7 @@
 <p align="center">
 <a href="https://packagist.org/packages/owen-it/laravel-auditing"><img src="https://poser.pugx.org/owen-it/laravel-auditing/v/stable.svg" alt="Latest Stable Version"></a>
 <a href="https://packagist.org/packages/owen-it/laravel-auditing"><img src="https://poser.pugx.org/owen-it/laravel-auditing/v/unstable.svg" alt="Unstable Version"></a>
-<a href="http://laravel.com"><img src="https://img.shields.io/badge/Laravel-5.2.x-brightgreen.svg?style=flat-square" alt="Laravel 5.2.x"></a>
-<a href="http://laravel.com"><img src="https://img.shields.io/badge/Laravel-5.3.x-brightgreen.svg?style=flat-square" alt="Laravel 5.3.x"></a>
-<a href="http://laravel.com"><img src="https://img.shields.io/badge/Laravel-5.4.x-brightgreen.svg?style=flat-square" alt="Laravel 5.4.x"></a>
+<a href="http://laravel.com"><img src="https://img.shields.io/badge/Laravel-5.x-brightgreen.svg?style=flat-square" alt="Laravel 5.x"></a>
 <a href="https://packagist.org/packages/owen-it/laravel-auditing"><img src="https://poser.pugx.org/owen-it/laravel-auditing/d/total.svg" alt="Total Downloads"></a>
 <a href="https://packagist.org/packages/owen-it/laravel-auditing"><img src="https://poser.pugx.org/owen-it/laravel-auditing/license.svg" alt="License"></a>
 
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,8 @@
         "php": ">=5.5.9",
         "illuminate/console": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
         "illuminate/database": "5.2.* || 5.3.* || 5.4.* || 5.5.*",
-        "illuminate/filesystem": "5.2.* || 5.3.* || 5.4.* || 5.5.*"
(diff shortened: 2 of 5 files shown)


quetzyg:
Hi @ymslavov,

Can you provide more information on this? I can tell it's a CSV driver, but what exactly is it trying to solve? What's the use case you have that requires this? I'm asking this, because a rollback feature would be a PITA to implement using CSV files.

ymslavov:
Hi @quetzyg,

This was purely written in good faith and willingness to progress the package forward. I believe it's something people will start to appreciate more in the following year with the introduction of the GDPR in the EU and its requirement for preserving an auditing trail in all systems dealing with private user data.

There isn't a particular business case behind it, rather an attempt to provide users with alternatives for storing the audit trail in other formats. Writing this, I assume this can also be implemented as a core method of the Auditor or as part of the Auditable trait (e.g. $model->toCsv($disk, $destination)) which can be combined with a scheduled task to export the db records to text files as sort of a back up or log.

All in all, happy to help with any tasks that may be outstanding, so give me a shout if you need anything done.

quetzyg:
Sounds good, I like it. @anteriovieira, what do you reckon?

anteriovieira:
@ymslavov thanks for the contribution, really awesome. Can you provide us with the documentation?

anteriovieira:
@quetzyg we can launch this feature in the next version of the package with php 7+. Do you have any thoughts on this?

quetzyg:
@anteriovieira, this doesn't introduce any breaking change, so it could potentially be part of a 4.x version.

I'll try to get some time to properly evaluate this and the other pull requests we have pending. This week has been busy for me.

ymslavov:
@quetzyg this driver requires league/csv:^9.0 which unfortunately requires php7.0. My understanding is 4.* still supports php5.5 and php5.6?

quetzyg:
@ymslavov, that's one thing I would like to get rid of. I'll try to make a code review ASAP.

ymslavov:
hey @quetzyg have you had time to review the changes yet?

quetzyg:
Sorry @ymslavov, not yet. Busy with work and family life.

I do have some concerns, though. How do you handle attributes being stored in a CSV format when they get added/removed from the model?

Imagine a `User` model with two properties: `name` and `surname`.

When something changes you get a CSV entry like:

```csv
new_name, new_surname, old_name, old_surname
Jon, , Jonny, 
```

But if you add a new attribute (let's say, email), you would then get:

```csv
new_name, new_surname, new_email, old_name, old_surname, old_email
Jon, , jon@example.com, Jonny, ,
```

But because the file already exists, you'll be appending, so the end result will be:

```csv
new_name, new_surname, old_name, old_surname
Jon, , jon@example.com, Jonny, ,
```

Which does not make sense.

Also, how do you differentiate an attribute that has not changed, from an attribute that was set to empty string?

As you can see, there are a few things that would have to be solved to make this CSV approach usable.

Let me know what you think.

ymslavov:
Hey @quetzyg ,

Currently the new_values and old_values are converted to json and serialized into two columns - New Values and Old Values. So varying number of updated fields between audits is not a concern at this point. The current implementation is a 1:1 version of the table that's being created with the database driver. Let me know if you have any concerns about this.

Another outstanding thing I wanted to discuss was pruning. Currently the prune() method just returns false as seeking inside a large audit file can have severe performance penalties. Do you think introducing a switch in the driver config, something like 'use_pruning' => true/false, would be a good idea? This way, if you know your system doesn't produce large audit files you can leave it on. Default would be false though.

quetzyg:
Hi @ymslavov,

Since I did not try the code, I understood it in another way. Can you show me an example CSV file with one or two lines, then?

ymslavov:
Sure @quetzyg , here is an example of two sequential operations on a model:

`
"Old Values","New Values",Event,"Auditable Id","Auditable Type","User Id",Url,"Ip Address","User Agent","Created At"
"{""is_primary"":1}","{""is_primary"":0}",updated,1,"App\Models\Address",4,http://proj.dev/api/user/1/address/1,172.21.0.1,"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36","2017-10-22 20:35:29"
"{""is_primary"":0}","{""is_primary"":1}",updated,1,"App\Models\Address",4,http://proj.dev/api/user/1/address/1,172.21.0.1,"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36","2017-10-22 20:35:29"
`

quetzyg:
Thanks for the sample @ymslavov.

So my question now is, what's the point of being a CSV if part of it will be in JSON format? You might as well output the whole object as JSON or XML. That would make more sense, than having two distinct formats mixed.

A flat file is good for structures that never change, which is not the case here.

@anteriovieira, thoughts?

ymslavov:
Hi @quetzyg ,

I partly agree with you that mixing both formats can be a bit weird. However, separating the rest of the data (apart from old_values and new_values) into separate columns provides the user with a convenient way to easily filter entries based on model ID for example.

Dumping the whole object as JSON would require a separate tool to parse all the entries and defies the initial idea of providing an easily accessible place to quickly check the audit trail, something like a log file.

quetzyg:
>Dumping the whole object as JSON would require a separate tool to parse all the entries

This implementation will delay the need of such tool, but you'll eventually have to use it, once you need to parse the attribute changes.

Also, what do you think will happen when an attribute for a textfield (think multiline) changes? Will the CSV parser break, because one row is suddenly split into two or three?

ymslavov:
I didn't get the last remark if I've to be honest, but if you think it doesn't align well with the package's core goals, then I'm willing to spin it up as a separate driver package for laravel-auditing. You can then decide to (or not to) endorse it in the docs. Let me know what you think.

quetzyg:
>Also, what do you think will happen when an attribute for a textfield (think multiline) changes? Will the CSV parser break, because one row is suddenly split into two or three?

Basically, what I mean by this is. If you have an attribute of the type `TEXT` (multi line), that will totally break the one line per audit CSV.

ymslavov:
Ah, ok, got it. Well making sure to escape all new lines after JSON.stringify()-ing the new_values and old_values objects would take care of it. So if you had \n as in a multiline text, we'd need to do `JSON.stringify(new_values).replace(/\\n/g, "\\n")` (which I currently don't do, so that's a totally valid point). Definitely a todo.

ymslavov:
@quetzyg @anteriovieira I've decided to spin this off as a separate driver package for laravel-auditing. It's located here: https://github.com/betapeak/laravel-auditing-filesystem

Feel free to close this PR and also I'd be grateful if you could reference the driver in the docs as an optional dependency if you like.

quetzyg:
Sounds like a plan, @ymslavov.

PR #340, opened 2017-11-14
Update Auditable.php

vyacheslav-kara:


ymslavov:
Looks good and makes sense for this to be configurable. Please just fix the styling.

@quetzyg what do you think?

quetzyg:
Thanks for your PR @vyacheslav-kara, I can't say why I didn't think of this before...

I will ask you, however, to do a PR against the `master` branch, please.

Also, update the configuration file with a new entry (below the `user` stuff):
```php
    /*
    |--------------------------------------------------------------------------
    | Auditable events
    |--------------------------------------------------------------------------
    |
    | The Eloquent events that should trigger an Audit.
    |
    */

    'events' => [
        'created',
        'updated',
        'deleted',
        'restored',
    ],
```

Also, pass one value per line:
```php
return Config::get('audit.events', [
        'created',
        'updated',
        'deleted',
        'restored',
]);
```

PR #378, opened 2018-01-29
Correct resolve user agent if user agent is null

MammutAlex:


quetzyg:
Thank for the PR @MammutAlex,

You're correct, there's already a pull request for this issue in the upcoming version.
The changes aren't that many from 5.0 to 6.0, the version bump was mainly because the configuration structure changed slightly and now we have proper resolvers that can be replaced, but the database structure remains the same.

I guess we can correct this in this version, but I'll do it differently.

Thanks, though.

Issue #310, opened 2017-09-20
Wrong format date when created


rorc:
| Q                 | A
| ----------------- | ---
| Bug?              | no
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.5.3
| Package version   | 4.1.2
| PHP version       | 7.0.1

#### Actual Behaviour
Saving date field with format `Y-m-d 00:00:00`

#### Expected Behaviour
Saving date field with format `Y-m-d`

#### Steps to Reproduce
Create any record, having date field.

quetzyg:
This doesn't seem to be an issue with the package.

rorc:
MySQL field type is `date`,
In model `$dates = ['date']`;
```php
class AuditingListener
{
    public function handle(Auditing $event)
    {
        dump($event->model->date); //already wrong value 2017-01-10 00:00:00
    }
}

Order::create([
    'date' => Request::input('date'); //2017-01-10
]);
```
After this, `new_values` has `2017-01-10 00:00:00`

quetzyg:
That's an Eloquent issue, which has been documented in the [troubleshooting](http://www.laravel-auditing.com/docs/4.1/troubleshooting) section.

rorc:
Can we solve this problem automatically at core level?
Before or after calling method `transformAudit` run code like that:
```php
if (isset($this->casts[$key]) && $this->casts[$key] == 'date') {
    return Carbon::parse($value)->toDateString();
}
```

rorc:
Now have to use many crutches before save audit, for example:
```php
if (isset($this->casts[$key]) && $this->casts[$key] == 'date') {
    return Carbon::parse($value)->toDateString();
}
```
And before output, for example:
```php
if ($value instanceof \DateTime && $value->toTimeString() == '00:00:00') {
    return $value->toDateString();
}
```

quetzyg:
Sorry, I'm not gonna fix upstream issues here. If they say it's an expected behaviour, we should keep it consistent. Bump the upstream issue and propose a fix there, not here.

rorc:
Ok ) Thank you for your sympathy )

PR #421, opened 2018-05-10
[FEAT] V7 - Multi user audits

quetzyg:
### Description
The code in this pull request is based on #420 but with a few simplifications.

**Work done:**
- [X] Audits are now related to Users via `MorphTo`;
- [X] UserResolver now checks multiple guards to resolve Users;
- [X] Updated migration with new `user_type` column;
- [X] Updated Audit factory;
- [X] Updated tests;
- [X] Other minor cleanups;

Special thanks to [Webparking](https://github.com/Webparking) for their initial work!

@crashkonijn & @remkobrenters

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -30,14 +30,18 @@ return [
     | User Keys, Model
     |--------------------------------------------------------------------------
     |
-    | Define the User primary key, foreign key and Eloquent model.
+    | Define the User Eloquent model, morph prefix and authentication guards
+    | to use in the User resolver.
     |
     */
 
     'user' => [
-        'primary_key' => 'id',
-        'foreign_key' => 'user_id',
-        'model'       => App\User::class,
+        'model'        => App\User::class,
+        'morph_prefix' => 'user',
+        'guards'       => [
+            'web',
+            'api',
+        ],
     ],
 
     /*
--- a/database/migrations/audits.stub
+++ b/database/migrations/audits.stub
@@ -27,7 +27,7 @@ class CreateAuditsTable extends Migration
     {
         Schema::create('audits', function (Blueprint $table) {
             $table->increments('id');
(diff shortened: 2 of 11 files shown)


crashkonijn:
Awesome job, thanks!

quetzyg:
@crashkonijn I should be tagging this in the next couple of days. Meanwhile, give it a go, to see if it works as advertised :smile:

crashkonijn:
@quetzyg I replaced our own changes with your changes this morning and everything works perfectly!

I take it you're going to tag it with version 7, since these changes break backwards compatibility?

Also, we just made a trait which fixes these issues: https://github.com/owen-it/laravel-auditing/issues/310 by copying the fix for Laravel 5.5. Would it be an idea to include it in the package and let people add it to their models when using Laravel 5.4 or earlier?

quetzyg:
I'm not a fan of fixing upstream issues downstream. Have you considered making a pull request to https://github.com/laravel/framework/ ?

But do share the trait, we might consider including it.

crashkonijn:
Do they still accept PR's on versions that old?

And I understand your point, but at the same time it does have a huge impact on the performance of this package (read: almost unusable) on older Laravel versions. That's why I suggested to only include it for people to include it if they need to.

Another option could of course also be to link to this gist in the docs: https://gist.github.com/crashkonijn/7d581e55770d2379494067d8b0ce0f6d :)

quetzyg:
>Do they still accept PR's on versions that old?

Try it, it wouldn't hurt, I guess :)

I'll have a look at the gist. :+1:

quetzyg:
@crashkonijn https://github.com/owen-it/laravel-auditing-doc/pull/62/files#diff-73654fb858684160f86a6b76afb2a6fdR47 :wink:

crashkonijn:
@quetzyg Awesome, thanks!

PR #429, opened 2018-06-13
Add jsonable on migration

lloricode:


quetzyg:
Sorry, but we don't want this.

PR #436, opened 2018-07-08
Globally Disable Auditing

rankarpan:


rankarpan:
@quetzyg There are cases where we don't want to audit while testing, running database seeding, etc... There needs to be a way to turn auditing off and on from configuration.

quetzyg:
You can already do [that](http://www.laravel-auditing.com/docs/7.0/auditable-configuration).

crashkonijn:
@quetzyg Are you referring to this?

```<?php
// Disable auditing from this point on
Article::disableAuditing();

// This operation won't be audited
Article::create([
    // ...
]);

// Re-enable auditing
Article::enableAuditing();```

I would also like to see the PR of @rankarpan merged because having a global switch for the audits is much easier than turning off auditing for every model separately during tests and/or dev seeding for example.

**Comment by quetzyg** [comment]
@crashkonijn yes I am. If testing and seeding are a concern, just keep the console logging disabled (default).

**Comment by crashkonijn** [comment]
@quetzyg We can't do that, we need that feature (hence that it's turned on).

**Comment by quetzyg** [comment]
> @quetzyg We can't do that, we need that feature (hence that it's turned on).

So then, how does this PR solve your problem? Currently, setting `audit.console` to `false` only disables auditing in the CLI. This proposal disables auditing across the board.

Both approaches rely on a value stored in the configuration file.

If it's because this PR uses the `env()` helper, you can just edit the configuration and change from this:

```php
'console' => false,
```

to this:
```php
'console' => env('AUDIT_CONSOLE', false),
```
What am I missing?

quetzyg:
I'm all in favour for code reuse, but if you have projects that don't need auditing, why are you pulling the dependencies and then wanting to disable the package entirely?

If you don't need it, don't use it.

crashkonijn:
I agree with you, but there is no other easy way to disable this and still have it in our core. We made the decision that including this package by default, and then disabling it in the rare chance that we don't need it outways having to add it on all entities separately.

Anyhow, besides this edge case the more common case that I really want to disable auditing is during seeding and tests. Of which the latter can also be browser tests and thus not necessarily running from the commandline.

crashkonijn:
I agree with you, but there is no other easy way to disable this and still have it in our core. We made the decision that including this package by default, and then disabling it in the rare chance that we don't need it outways having to add it on all entities separately.

Anyhow, besides this edge case the more common case that I really want to disable auditing is during seeding and tests. Of which the latter can also be browser tests and thus not necessarily running from the command line.

PR #439, opened 2018-07-20
Add availability to nonScalar for recursive Array and noSQL usage

bsormagec:
If you are using NoSQL you need this. Because noSQL has ability to recursive array or objects and you can save them but in this package its returning null always changes. With this change you can enable it from config. Thanks.

`{
"id":"1",
"name": "test",
"surname": "user",
"settings": {
"hasCar": true,
"hasBike": false,
"isParent": false,
},
"isActive":true
}`

In this json example when i change settings object now this package not check changes. but with my changes if anybody wanna use this only enable it from settings.

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -23,7 +23,7 @@ return [
     |
     */
 
-    'implementation' => App\Models\Audit::class,
+    'implementation' => OwenIt\Auditing\Models\Audit::class,
 
     /*
     |--------------------------------------------------------------------------
@@ -105,7 +105,7 @@ return [
     |
     */
 
-    'threshold' => 100,
+    'threshold' => 0,
 
     /*
     |--------------------------------------------------------------------------
@@ -141,7 +141,7 @@ return [
     'drivers' => [
         'database' => [
             'table'      => 'audits',
-            'connection' => 'mongodb',
+            'connection' => null,
         ],
     ],
 
--- a/config/audit.php
+++ b/config/audit.php
@@ -23,7 +23,7 @@ return [
     |
     */
 
-    'implementation' => OwenIt\Auditing\Models\Audit::class,
+    'implementation' => App\Models\Audit::class,
 
     /*
     |--------------------------------------------------------------------------
@@ -36,7 +36,7 @@ return [
 
     'user' => [
         'morph_prefix' => 'user',
-        'guards'       => [
+        'guards' => [
             'web',
             'api',
         ],
@@ -51,10 +51,10 @@ return [
     |
     */
     'resolver' => [
-        'user'       => OwenIt\Auditing\Resolvers\UserResolver::class,
+        'user' => OwenIt\Auditing\Resolvers\UserResolver::class,
         'ip_address' => OwenIt\Auditing\Resolvers\IpAddressResolver::class,
         'user_agent' => OwenIt\Auditing\Resolvers\UserAgentResolver::class,
-        'url'        => OwenIt\Auditing\Resolvers\UrlResolver::class,
+        'url' => OwenIt\Auditing\Resolvers\UrlResolver::class,
     ],
 
     /*
@@ -105,7 +105,7 @@ return [
     |
     */
 
-    'threshold' => 0,
+    'threshold' => 100,
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


config/audit.php
```diff
@@ -23,7 +23,7 @@
     |
     */
 
-    'implementation' => OwenIt\Auditing\Models\Audit::class,
+    'implementation' => App\Models\Audit::class,
```
quetzyg: Why was this changed?

config/audit.php
```diff
@@ -36,7 +36,7 @@
 
     'user' => [
         'morph_prefix' => 'user',
-        'guards'       => [
+        'guards' => [
```
quetzyg: Lets leave the array alignment alone, please.

config/audit.php
```diff
@@ -105,7 +105,7 @@
     |
     */
 
-    'threshold' => 0,
+    'threshold' => 100,
```
quetzyg: Put it back as it was.

config/audit.php
```diff
@@ -140,8 +140,8 @@
 
     'drivers' => [
         'database' => [
-            'table'      => 'audits',
```
quetzyg: Change it back.

config/audit.php
```diff
@@ -155,4 +155,16 @@
...
+    /*
+    |--------------------------------------------------------------------------
+    | Audit Allow Saving NonScalar Values and Recursive Arrays
+    |--------------------------------------------------------------------------
+    |
+    | Allow Saving NonScalar Values and Recursive Arrays
+    |
+    */
+
+    'nonScalar' => true,
```
quetzyg: Follow the configuration index convention and use `snake_case`.

config/audit.php
```diff
@@ -141,7 +141,7 @@
     'drivers' => [
         'database' => [
             'table'      => 'audits',
-            'connection' => null,
+            'connection' => 'mongodb',
```
quetzyg: Put it back.

config/audit.php
```diff
@@ -155,4 +155,15 @@
     */
 
     'console' => false,
+
+    /*
+    |--------------------------------------------------------------------------
```
quetzyg: What's with the double header?

config/audit.php
bsormagec: Im fixing all back 🐙

src/Auditable.php
```diff
@@ -103,13 +103,15 @@ protected function resolveAuditExclusions()
...
-            // Apart from null, non scalar values will be excluded
-            if (is_array($value) || (is_object($value) && !method_exists($value, '__toString'))) {
-                $this->excludedAttributes[] = $attribute;
+        if (!Config::get('audit.nonScalar', false)) {
+            // Valid attributes are all those that made it out of the exclusion array
+            $attributes = array_except($this->attributes, $this->excludedAttributes);
+
+            foreach ($attributes as $attribute => $value) {
+                // Apart from null, non scalar values will be excluded
+                if (\is_array($value) || (\is_object($value) && !method_exists($value, '__toString'))) {
```
quetzyg: Also, remove the backslash from `is_array()` and `is_object()`.

bsormagec: Sure...

config/audit.php
bsormagec: fixed

config/audit.php
bsormagec: fixed

config/audit.php
bsormagec: fixed

config/audit.php
bsormagec: fixed

quetzyg:
>If you are using NoSQL you need this. Because noSQL has ability to recursive array or objects and you can save them but in this package its returning null always changes. With this change you can enable it from config. Thanks.

Please provide an example of what the actual issue is, and how this change fixes it.

bsormagec:
`{
"id":"1",
"name": "test",
"surname": "user",
"settings": {
"hasCar": true,
"hasBike": false,
"isParent": false,
},
"isActive":true
}`

In this json example when i change settings object now this package not check changes. but with my changes if anybody wanna use this only enable it from settings.

quetzyg:
In order to go forward with this, you'll have to add tests for the use case you're describing.

Test to pass, test to fail, every scenario.

quetzyg:
Closing due to lack of updates

PR #445, opened 2018-09-16
Swaps Drop for DropIfExists

yakidahan:


quetzyg:
What's the reason?

yakidahan:
Well one use case would be when I want to rollback the migrations, but before that I delete the `audits` table by hand from the database. When using `drop` with the rollback you will get an error since the table doesn't exist anymore and the rollback would fail. If I use `dropIfExists` the rollback would succeed.

quetzyg:
Sorry, but I don't see the point. If you're deleting tables by hand, you might as well delete the appropriate entry from the `migrations` table too, while you're at it (if you're into manually changing stuff).

The point of using migrations is that you don't make changes by hand, you just migrate/rollback via artisan.

yakidahan:
In addition it is the default of laravel migration: [https://github.com/laravel/framework/blob/master/src/Illuminate/Database/Migrations/stubs/create.stub](https://github.com/laravel/framework/blob/master/src/Illuminate/Database/Migrations/stubs/create.stub)

quetzyg:
If you have that specific need, do the change on the applications using this package. Like stated in the [documentation](http://www.laravel-auditing.com/docs/8.0/audit-migration), there's plenty of customisations that you can perform in your migration, but they don't necessarily have to be included by default.

PR #462, opened 2018-11-14
Add ability to short circuit trait booting

dallincoons:
This is a very simple, easy tweak. The goal is to allow for doing things like testing individual methods on an Eloquent Model without needing to load the framework, since running a test that boots the Laravel framework is typically three times slower than one that doesn't. 

Because `bootAuditable` immediately calls `isAuditingEnabled`, which references the `App` facade, you can't instantiate a model in a Laravel-free test without getting an error having to do with the facade not being available. 

This tweak allows the user to simply add `Article::$auditingDisabled = true;` to the beginning of the test, and the problem goes away.


cwilby:
Missed this the first time around on this issue, thanks @dallincoons

PR #467, opened 2018-11-22
Request to fix #466

### LINKED ISSUE #466: Migration stub inconsistant causing migration errors
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.7
| Package version   | 8.0.3
| PHP version       | 7.1

#### Actual Behaviour

Running `artisan vendor:publish --tag=migration` generates migration file with timestamp prefix.
Run the command again and it generates a new file with a different file name but the same class name.

#### Expected Behaviour

Describe the behaviour you're expecting.

Running `artisan vendor:publish --tag=migration` generates migration file with timestamp prefix of when it was first created (should be around the time this package was made public).

Running it again should not create new file and contents should only update if the `force` flag is used.

The above means no other package using the `--tag=migration` tag will work, ever. It will keep generating new file and getting error `Cannot declare class CreateAuditsTable, because the name is already in use` on publishing migrations from other packages.

#### Steps to Reproduce

see above

#### Possible Solutions

Take the timestamp generation off the publish method.

lukesnowden:
Use the current filename if migration exists, if not, use a static filename from now on.

### DIFF
--- a/src/AuditingServiceProvider.php
+++ b/src/AuditingServiceProvider.php
@@ -5,6 +5,10 @@ namespace OwenIt\Auditing;
 use Illuminate\Support\ServiceProvider;
 use OwenIt\Auditing\Console\AuditDriverMakeCommand;
 use OwenIt\Auditing\Contracts\Auditor;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+use RecursiveRegexIterator;
+
 
 class AuditingServiceProvider extends ServiceProvider
 {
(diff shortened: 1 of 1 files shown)


lukesnowden:
@quetzyg is there any reason why this isn't being merged in? I have been using the above for the past 2 weeks with no issues.

quetzyg:
@lukesnowden, sorry, I've been busy with other stuff and forgot to comment.

My issue with this approach is the complexity involved for such a simple thing.

It's easier to rename the file from `audits.stub` to something like `2018_12_06_000001_create_audits_table.php` and update the logic in the `AuditingServiceProvider` from 

```php
        $this->publishes([
            $migration => database_path(sprintf('migrations/%s_create_audits_table.php', date('Y_m_d_His'))),
        ], 'migrations');
```

to
```php
        $this->publishes([
            $migrations => database_path('migrations'),
        ], 'migrations');
```

quetzyg:
And then, there's the coding style, too.

Issue #486, opened 2019-02-08
Version 8.0.3 breaks support for non-integer keys 


specialtactics:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.7.25
| Package version   | >=8.0.3
| PHP version       | 7.2.x

#### Actual Behaviour

This PR here
https://github.com/owen-it/laravel-auditing/pull/448/files

For this issue (from what I understand):
https://github.com/owen-it/laravel-auditing/issues/432

Will actually break any project using non-numeric (or non-integer) primary key IDs. In our case, we are using UUIDs (which I would think is quite common these days). 

The end result is that any update to the underlying model will wipe all audits and it will no longer have any related audits.

#### Expected Behaviour

As laravel supports non-integer keys, so should this package. Update shouldn't break existing functionality.

#### Steps to Reproduce

Create a new model in Laravel with the following properties and as a string in the database (for field type)

```php
    /**
     * @var bool Set to false for UUID keys
     */
    public $incrementing = false;

    /**
     * @var string Set to string for UUID keys
     */
    protected $keyType = 'string';
```

Deploy the auditable trait on it, and see what happens on model update.

#### Possible Solutions

The auditable_id cast to integer needs to be removed

jonagoldman:
You can also extend the Audit model and add uuid behaviour:

```php
<?php

namespace App\Models;

use Illuminate\Support\Str;
use Illuminate\Database\Eloquent\Model;
use OwenIt\Auditing\Contracts\Audit as AuditContract;
use OwenIt\Auditing\Audit as BaseAudit;

class Audit extends Model implements AuditContract
{
    use BaseAudit;

    /**
     * The "booting" method of the model.
     *
     * @return void
     */
    protected static function boot()
    {
        parent::boot();

        static::creating(function($model) {

            if (is_null($model->getAttribute('id')))
            {
                $model->id = Str::orderedUuid();
            }
            
        });

    }

    /**
     * The "type" of the auto-incrementing ID.
     *
     * @var string
     */
    protected $keyType = 'string';

    /**
     * Indicates if the IDs are auto-incrementing.
     *
     * @var bool
     */
    public $incrementing = false;

    /**
     * {@inheritdoc}
     */
    protected $guarded = [];

    /**
     * {@inheritdoc}
     */
    protected $casts = [
        'id' => 'string',
        /// ... others
    ];

}

```

[...]

PR #487, opened 2019-02-08
Fix for issue #486 - 

specialtactics:
remove auditable_id integer cast, to allow for non-integer primary key models

### DIFF
--- a/src/Models/Audit.php
+++ b/src/Models/Audit.php
@@ -19,6 +19,6 @@ class Audit extends Model implements \OwenIt\Auditing\Contracts\Audit
     protected $casts = [
         'old_values'   => 'json',
         'new_values'   => 'json',
-        'auditable_id' => 'integer',
+        // Note: Please do not add 'auditable_id' in here, as it will break non-integer PK models
     ];
 }


specialtactics:
Wondering if I can get any comments about this from you @squiaios, it's blocking our upgrade to Laravel 5.8 (and I'm sure we can't be the only ones).

src/Models/Audit.php
```diff
@@ -19,6 +19,6 @@ class Audit extends Model implements \OwenIt\Auditing\Contracts\Audit
     protected $casts = [
         'old_values'   => 'json',
         'new_values'   => 'json',
-        'auditable_id' => 'integer',
+        // Note: Please do not add 'auditable_id' in here, as it will break non-integer PK models
```
Namoshek: Why does this break non-integer PKs? If you override the model in your code to use another type of PK, you can also override the `$casts`. In my opinion, this should not be changed as it would force others (like me) that are using the default configuration to add this cast manually in a customized model.

specialtactics: @Namoshek This is a vendor model, you can't just override it. The reason this breaks non-integer PKs is because it casts UUID PKs (or any stirng PKs) as integer, thereby changing them completely. I added a test in my PR which demonstrates this.

The cast shouldn't be necessary in any case - because auditable_id could be of any type in theory (but in particular integer or string), so this package shouldn't make assumptions as to it's type - especially given Laravel/Eloquent itself supports string key types directly.

Namoshek: Of course you can override it. It is the first option in the config:

https://github.com/owen-it/laravel-auditing/blob/eff1fe27f3be42eeafbd72ffb9ea55b37a3142be/config/audit.php#L5-L14

specialtactics: @Namoshek So can you explain why someone would need to override this model for acquire it's compatibility with default Laravel functionality?

If you have some special case where it doesn't work for you, then perhaps you should override it yourself, but the way it is in my PR works for all keys of all major databases (Postgres, MySQL, MariaDB, etc).

Namoshek: Yes, I do have a reasoning why it should be an integer cast:
- The default migration is for an integer column. If you adjust the migration, you can or should also adjust the model.
- The default Eloquent behavior is an integer primary key.
- The majority of users will use integer primary keys throughout their whole application, making it the obvious choice for the more suitable default.

In the end it is the decision of the maintainer. He did announce the change in the changelog, although I have to admit that it should have been a minor release as it may break BC as in your case.

specialtactics: @Namoshek The fact that it's "default" makes no difference, as both are equally supported functionality, and it will work for both without casting.

You can not go making assumptions what "the majority" of users are using, and it is not relevant anyway - so long as it's a core feature of Laravel, this package shouldn't break that.

PR #489, opened 2019-02-25
Laravel 5.8 support

squiaios:


delenamalan:
Unfortunately it's not this simple to support Laravel 5.8. Have a look at the [Upgrade Guide](https://laravel.com/docs/5.8/upgrade). There are a few things that have been deprecated that this package still supports. For example, this package still uses the deprecated [string and array helpers](https://laravel.com/docs/5.8/upgrade#string-and-array-helpers), the `defer` property of [service providers](https://laravel.com/docs/5.8/upgrade#deferred-service-providers) and the deprecated `fire` method of the `Illuminate/Events/Dispatcher` class.

All of the packages that this package requires also need to support 5.8, but [`orchestral/database` ](https://packagist.org/packages/orchestra/database)doesn't yet. :disappointed:

squiaios:
@delenamalan thanks, I fixed all the issues you mentioned
orchestral database & testbench has version 3.8, so that is not a problem

delenamalan:
[orchestra/database](https://packagist.org/packages/orchestra/database) 5.8 hasn't been released yet, but hopefully will be soon :+1:

crynobone:
Why does it matter for a require-dev to have stable deps, none of the user using the project will require it, the travis configuration does run on any of the previous Laravel build except latest stable and orchestra/database is no longer required since orchestra/testbench 3.6

composer.json
```diff
@@ -39,16 +39,16 @@
...
+        "illuminate/filesystem": "^5.8"
     },
     "require-dev": {
-        "phpunit/phpunit": "^6.0",
+        "phpunit/phpunit": "^7.5",
         "mockery/mockery": "^1.0",
-        "orchestra/testbench": "^3.5",
-        "orchestra/database": "^3.5"
+        "orchestra/testbench": "^3.8",
+        "orchestra/database": "^3.8"
```
crynobone: `orchestra/testbench` `3.8` doesn't needs `orchestra/database`, you can simply remove the deps and service provider in your tests.

squiaios: @crynobone thanks for your advices, I removed what you said :)

squiaios:
I fixed composer deps and unit test

anteriovieira:
Thank you for your contribution @squiaios

squiaios:
> Thank you for your contribution @squiaios

with pleasure @anteriovieira

vpillinger:
Are there any changes required when going from 8.0.0 to 9.0.0? 

Either way, it should probably be mentioned here: http://www.laravel-auditing.com/docs/9.0/upgrading

Issue #435, opened 2018-07-08
Globally disable auditing


rankarpan:
There are cases where we don't want to audit while testing, running migrations, etc... There needs to be a way to turn auditing off and on.

andreshg112:
Is there a way to disable in the config file or .env? @quetzyg

quetzyg:
Hi @andreshg112, 

If you mean disabling the auditing functionality entirely, no, this package doesn't allow for that.

Refer to the [Auditable Configuration](http://www.laravel-auditing.com/docs/9.0/auditable-configuration), more specifically the `Enable/Disable` section.

**PS:** I'm no longer a collaborator here, since I started my own auditing package.

PR #497, opened 2019-03-29
Added audit.enabled config.

andreshg112:
This allows disabling auditing globally using the config:
`'enabled' => env('AUDITING_ENABLED', true),`

It "fixes" https://github.com/owen-it/laravel-auditing/issues/435

I decided to implement it because I'm using a custom driver that sends notifications through AWS SNS, so I need to disable auditing locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 /vendor
 composer.lock
+.vscode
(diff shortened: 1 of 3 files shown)


andreshg112:
@anteriovieira , is it possible to merge this? Sorry for tagging you.

PR #550, opened 2019-09-13
Bugfix/fix for non integer key models

specialtactics:
remove auditable_id integer cast, to allow for non-integer primary key models

Original PR described here: https://github.com/owen-it/laravel-auditing/pull/487

It fixes a bug introduced which breaks default implementation of this package for non-integer primary keys


anteriovieira:
Thanks @specialtactics , fixed in the version `v9.3.2`

specialtactics:
Thank you @anteriovieira !

PR #557, opened 2019-10-23
Remove none need package

lloricode:
Remove none need package, then add `.phpunit.result.cache` to `.gitignore`


lloricode:
https://github.com/owen-it/laravel-auditing/pull/560#pullrequestreview-309550854

PR #545, opened 2019-09-03
Laravel 6

divdax:
Support Laravel 6 😄

PR #560, opened 2019-10-30
Substitute str_replace_last with the respective Str helper

koernchen02:
to remove the seemingly last occurence of the now obsolete str_* methods and with that the dependency on laravel/helpers package

as mentioned at #545


lloricode:
hi, forgot to mention, laravel/helpers suggest to remove to prevent use this in projects

PR #574, opened 2020-02-14
Miss dev requirement

raphaeldealmeida:
Sqlite3 e extension is required to run tests. Is interesting explicit require them in dev mode

sqlite3 is used here
https://github.com/owen-it/laravel-auditing/blob/master/tests/AuditingTestCase.php#L22


MortenDHansen:
Thanks 👍  - included in upcoming minor

PR #577, opened 2020-03-03
laravel 7

recca0120:


MrReeds:
Namespace detection should be still allowed, as some laravel 7 implementations are upgrades from past versions that still supported renaming the namespace by default; also the namespace renaming is still available trough third party and should not be hardcoded anyway

recca0120:
https://github.com/laravel/framework/commit/100698ea302b96d3fb251ccf6e63f6baaca25c00

in this commit, laravel 7 already remove DetectsApplicationNamespace
if you need this, it still can get namespace from Application
so I update this pr.

MrReeds:
Thank you, this seems more logical solution.

PR #583, opened 2020-03-05
Fix: Disable audits with config even if is running on console

marcoocram:
Sometimes we need to disable audits even running on console (for example on tests) but without this to get this behaviour we have to set the config running on console to false.

With this fix, disable audits will be disabled on all environments.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -4,7 +4,6 @@ risky: true
 
 enabled:
   - php_unit_strict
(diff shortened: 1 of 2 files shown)


MortenDHansen (approved):

MortenDHansen:
Thanks for adding this. 👍

anteriovieira (approved):
Thanks!

PR #587, opened 2020-05-24
Update audits.stub

lesichkovm:
The change solves the package being able to store the user id and and the auditable id which use unique IDs (UUID, GUID, etc)


MortenDHansen:
Thank you for submitting! 
The package provides a default that matches an out-of-the-box setup. The preferred procedure for supporting other than the default is to make the adjustments as needed. See http://www.laravel-auditing.com/docs/9.0/audit-migration

PR #594, opened 2020-07-15
Resolve #263 by making storage of empty values optional

### LINKED ISSUE #263: Old and new values are both empty JSON arrays
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.3.31
| Package version   | 4.0.7
| PHP version       | 7.1.6

#### Actual Behaviour

The `old_values` and `new_values` are both empty JSON arrays `[]`.

#### Expected Behaviour

No audit entry when new and old values are both empty.

#### Steps to Reproduce

Set console auditing to true and add `$auditExclude` columns, then only update those excluded columns thru the model.

#### Possible Solutions

In `OwenIt\Auditing\Drivers\Database` change this method:
```
public function audit(Auditable $model)
{
    return Audit::create($model->toAudit());
}
```

...to this:
```
public function audit(Auditable $model)
{
    if (empty($model->toAudit()['old_values']) && empty($model->toAudit()['new_values'])) {
        return null;
    }
    return Audit::create($model->toAudit());
}
```

DigitalMachinist:
I wrote this fix in order to resolve issue #263 in the course of integrating this package with a project at work.

It seems as though there's opinions regarding whether its valuable to store audits for changes that result in empty `old_values` & `new_values` or not, and in the case of my team, we'd prefer not to since storing all those extra records entails a lot more DB writes. Unfortunately, writing a different driver was not enough to allow me to resolve this because there's no way to return `null` from the Driver's `audit()` method to indicate that you don't want to store anything.

The workaround offered by @quetzyg works but still results in wasted computation creating the Audit record just to throw it away.

I'm submitting this to provide a way for users to choose how to handle audits when both the `old_values` & `new_values` records are empty, and additionally making it possible for people to return `null` from a driver to indicate for any reason that they do not want to store an audit record. I've included tests to confirm that this behaviour is working as intended.

It's my first PR submission so please let me know how I can improve upon this!


src/Auditor.php
```diff
@@ -66,10 +66,13 @@ public function execute(Auditable $model)
             return;
         }
 
-        if ($audit = $driver->audit($model)) {
-            $driver->prune($model);
+        $audit = $driver->audit($model);
+        if (!$audit) {
+            return;
```
DigitalMachinist, line 71: If no record was returned by the driver, abort the audit.

src/Drivers/Database.php
```diff
@@ -16,6 +16,15 @@ public function audit(Auditable $model): Audit
...
         $implementation = Config::get('audit.implementation', \OwenIt\Auditing\Models\Audit::class);
 
+        // If we want to avoid storing Audits with empty old_values & new_values, return null here.
+        if (!Config::get('audit.empty_values')) {
+            $attributeGetter = $model->resolveAttributeGetter($model->getAuditEvent());
+            list($old, $new) = $model->$attributeGetter();
+            if (empty($old) && empty($new)) {
+                return null;
+            }
+        }
```
DigitalMachinist, line 26: I basically just stole the logic for getting the `$old` and `$new` from `Auditable` and it feels like there's probably a cleaner way to encapsulate this. Open to suggestions.

anteriovieira:
Hi @DigitalMachinist, sorry for the delay, can you fix the conflict?

MortenDHansen:
I have pulled the implementation into upcoming v13 release, but made a few modifications so that we don't break backwards compatibility with existing drivers and we still support the "retrieved" event which is intentionally always empty.
(https://github.com/owen-it/laravel-auditing/pull/684)
Thank you for submitting!!

PR #605, opened 2020-09-07
Use Model's serializeDate method for formatting.

bskl:
When the Model override the serializeDate method for different date formatting, Audit keep using Elequent Model method. With calling $this->auditable->serializeDate($value) if avaliable, allows to using model's method.


MortenDHansen:
Thank you for making this PR 👍 
 - will be implemented shortly!

PR #607, opened 2020-09-08
Upgrade to Laravel 7

naszybko:
I've modified composer in order to make it compatible with Laravel 8 requirements.


jasonlewis:
The `Auditor` manager class still makes use of the now removed (previously deprecated) `$app` property. This needs to be changed to the `$container` property.

PR #612, opened 2020-09-14
L8 test

recca0120:
use orchestra/testbench 6.0 to test laravel 8 and add laravel 5.8 support


MortenDHansen:
Thanks for submitting! - real good stuff!
I've just updated the test suite with many of the same adjustments you made, but yours is slightly more elegant 😄

PR #631, opened 2021-01-14
Set the fallback value of auditable_id column to 0

mo7zayed:
The case is having an auditable pivot model with `id` column.

Deleting the pivot record causes an issue 
```txt
SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'auditable_id' cannot be null (SQL: insert into `audits` (`old_values`, `new_values`, `event`, `auditable_id`, `auditable_type`, `user_id`, `user_type`, `url`, `ip_address`, `user_agent`, `tags`, `updated_at`, `created_at`)
```

The solution was to set a fallback value for the `auditable_id` column to 0


MortenDHansen:
auditable_id cannot be null, functionality is built against auditable. Version 13 has support for a way to implement pivot auditing on the parent auditable (pr #684). I don't want to lift this requirement without analyzing and covering the change with tests.

PR #637, opened 2021-03-17
set define auditExclude in config

ShahbozHalim:


MortenDHansen (approved):

MortenDHansen:
Thanks for the contribution! - we will merge it, but i'm just considering how this would work the best 

- As suggested here, it will work like "audit_exclude_defaults". If nothing is defined, it will take these values. 
- Maybe it would be better if it works like a merge so that it merges the global excludes with model specific excludes. 
- Finally it could be default excludes which can be overwritten on a model level. (which would require to make it possible to opt out of excludes like with eloquents $guarded = [] array

MortenDHansen:
Needed to refactor the exclusion logic to make it play nice with local overrides. It will be included with the v13 release

PR #644, opened 2021-04-18
do not publish migration if exists

corbancode:
This PR fixes publishing the migration file everytime vendor publish is done.


MortenDHansen:
A very good idea! 
I think it's slightly cleaner (for now at least) to just check for class existence until we adopt anonymous migrations. (https://github.com/owen-it/laravel-auditing/pull/692)

PR #650, opened 2021-05-24
Correcting the observer when the model implements more Classes

rcarvs:
When the app implements more than the Audit Contract class, the observer of Audit tries to retrieve these other classes too, giving a type error. An example of a class that generates the error:
```
class Upload extends Model implements Auditable,HasMedia
{
```

To solve this I removed the type obligation of the method and add a condition to call the execute only if it is what we expect.

### DIFF
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -21,9 +21,11 @@ class AuditableObserver
      *
      * @return void
      */
-    public function retrieved(Auditable $model)
+    public function retrieved($model)
     {
-        Auditor::execute($model->setAuditEvent('retrieved'));
+        if($model instanceof Auditable){			
+            Auditor::execute($model->setAuditEvent('retrieved'));
+        }
     }
 
     /**


MortenDHansen:
Thank you for submitting!
I can't seem to reproduce the error. Can you help me describe how to provoke the error. I would like to create a test that exposes the problem

MortenDHansen:
Thank you for getting back! 😄 

So, i have tried it out doing the following:

Made a test case looking like this:
```
    /**
     * @return void
     * @test
     */
    public function observerHandlesMultipleInterfaces()
    {
        $model = factory(Article::class)->make();
        $observer = new AuditableObserver();
        $observer->retrieved($model);
        $this->assertTrue(true);
    }
```

On the Article model i have implemented two interfaces

```
<?php

namespace OwenIt\Auditing\Tests\Models;

use Illuminate\Database\Eloquent\SoftDeletes;
use OwenIt\Auditing\Contracts\Auditable;

class Article extends Model implements HasJediPowers, Auditable
{
```

(Thats the interface:)
```
<?php

namespace OwenIt\Auditing\Tests\Models;

interface HasJediPowers
{

    public function applyMindcontrol($subject);
}
```

Now, if i remove the implements Auditable - it barfs a TypeError like expected, but no amount of added interfaces will make it barf, as long as it has the Auditable.

Looking into the documentation, if using the Interface name as type for the argument in the observer, it will use functionality of instanceof (https://www.php.net/manual/en/language.types.declarations.php). 

Am i testing it wrongly?

MortenDHansen:
Reopen if relevant 🙂

Issue #567, opened 2020-01-07
Excluding a column still writes an entry even if there is no other data updated


fri3ndly:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.8.0
| Package version   | 9.0
| PHP version       | 7.3

#### Actual Behaviour

Added an exclusion column on model:
```
protected $auditExclude = [
        'last_active_at',
    ];

```

A blank data entry is added to the audit table e.g. old_values = [], new_values = []

#### Expected Behaviour

Expecting this to not make an entry in the database if no data is updated - Only an excluded column is updated

martin-arvisais:
Same here, any fix?

jefc1111:
Also interested in this. I just migrated to this from a different package after upgrading my app to Laravel 6. All is great apart from this point! Cheers.

tekook:
Still present in version 12.x

cwilby:
May be possible to add logic to this function

```
    public function audit(Auditable $model): Audit
    {
        $implementation = Config::get('audit.implementation', \OwenIt\Auditing\Models\Audit::class);

        $audit = $model->toAudit();

        if ($audit['new_values'] === []) {
           return;
        }

        return call_user_func([$implementation, 'create'], $model->toAudit());
    }
```

cwilby:
FYI I've created a PR to solve this issue.  If merged, you will soon be able to add `$createEmptyAudits = false` to your models to fix this issue.

MortenDHansen:
v13 has a new implementation of the excludes and this should now be supported: See https://www.laravel-auditing.com/docs/13.0/auditable-configuration "Empty audits"

PR #662, opened 2021-10-04
Add option to prevent creating empty audit records for excluded fields

cwilby:
Hello 👋

This PR addresses the issue found in https://github.com/owen-it/laravel-auditing/issues/567.  

Currently, if a model has an `auditExclude` array present, and an update that changes only fields in that array happens, an empty audit record is created.

These changes add a `createEmptyAudits` property to the Auditable trait that is **true** by default.

When this is set to **false**, additional logic is run to bail on insert in the correct scenario.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -209,6 +209,14 @@ trait Auditable
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function shouldCreateEmptyAudits(): bool
+    {
+        return true;
+    }
+
+    /**
      * Modify attribute value.
      *
      * @param string $attribute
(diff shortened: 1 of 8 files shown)


cwilby:
@anteriovieira It looks like it is mostly styleci issues with this branch.  Tests are passing locally, please advise.

MortenDHansen:
closes issue #567

MortenDHansen:
I have pulled the implementation into upcoming v13 release, but made a few modifications so that we don't break backwards compatibility with existing drivers and we still support the "retrieved" event which is intentionally always empty.

Thank you for submitting!!

Issue #670, opened 2021-12-01
[PHP 8.1] Deprecation message


robsontenorio:
| Q                 | A
| ----------------- | ---
| Bug?              | no
| New Feature?      | yes
| Framework         | Laravel
| Framework version | 8.73
| Package version   | 12.0
| PHP version       | 8.1

#### Actual Behaviour

```
preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /var/www/app/vendor/owen-it/laravel-auditing/src/Auditable.php on line 423
```

#### Expected Behaviour

No warning message.

#### Steps to Reproduce

Just save any model to fire auditing.

#### Possible Solutions

Type hint suggest methods as described on deprecation message.

repat:
Could you please fix @anteriovieira  :)

PR #671, opened 2021-12-06
Fix PHP 8.1 deprecation message

repat:
This created the deprecation message:

> PHP Deprecated:  preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /xyz/vendor/owen-it/laravel-auditing/src/Auditable.php on line 426

The PHPDoc says to return null but the code wouldn't have so I just put in an early return. Didn't have any problems with PHP 8.1 after that

See also #670

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -415,6 +415,10 @@ trait Auditable
      */
     protected function resolveAttributeGetter($event)
     {
+        if(empty($event)) {
+            return;
+        }
+
         foreach ($this->getAuditEvents() as $key => $value) {
             $auditableEvent = is_int($key) ? $value : $key;
 


robsontenorio:
For reference, another approach to solve it 

https://github.com/barryvdh/laravel-debugbar/pull/1249/files

Rian190:
Any progress on this one? We would like to update to 8.1 :)

repat:
@Rian190  you can use it with PHP 8.1, it just throws these annoying warnings

XGhozt:
+1

aglipanci:
It would be great if someone can merge this :)

MortenDHansen (approved):

MortenDHansen:
Thanks for the PR. Will be part of upcoming minor release

PR #673, opened 2022-01-26
Proposing changes to the Auditable and resolvers

MortenDHansen:
This PR suggests a few changes in the project.

1. Allowing for adding and removing resolvers
Apart from the user resolver, users of the package may or may not want to add or remove resolvers. 
examples: 
- In some applications the User-agent may be of no interest. 
- In some cases developer may want to add an additional column to Audit table and supply another meaningful property

With this change, resolvers listed in the config array "resolvers" are used instead of the previously hardcoded list of resolvers. Existing resolvers are still in the array, except the User resolver which is required. That has been moved to the user config key

2. Resolvers now receives an instance of the model as argument (passed by reference). 
3. "Contracts" has been cleaned. The Resolver contract is the same for all Resolvers (static method resolve).

Note that this is breaking (published config file needs to be updated)


MortenDHansen:
Closed for restructuring 😄

PR #682, opened 2022-02-02
Update default guards to match Laravel changes

falko189:
Update the default api guard to be sanctum following the update on laravel here
https://github.com/laravel/laravel/commit/226d1bfc3c774e24a79e68e9f1264b5ca359551f#diff-66a4eb6caa45de87ad409f2f5f2d5e7dd30aacd12de32af921aa32317cf9b2e2L45

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -28,7 +28,7 @@ return [
         'morph_prefix' => 'user',
         'guards'       => [
             'web',
-            'api',
+            'sanctum',
         ],
     ],
 


anteriovieira:
related https://github.com/laravel/laravel/pull/5663

falko189:
That's a fair point, we could release a major version with this as breaking change. But at the very least we should mention it in the Readme that this is a known issue 👍🏻  Let me know and I'll put together something

MortenDHansen:
Hi guys, just spitballing here, but could this be a solution?

https://github.com/owen-it/laravel-auditing/pull/695

Basically we skip the guard if it can't be resolved and for new installs we leave the config array empty. Upon running the user resolver it will fall back to the default configured in laravel auth.php

falko189:
Yep your solution is much more elegant and should cover old laravel versions :) I'll close this off! thanks

PR #684, opened 2022-02-04
V13 dev

MortenDHansen:
Start of v13

- Resolvers are now dynamic and can be used to fill whatever columns user may wish. 
- Default resolvers may be disabled
- It is possible to discard empty audits (new and old values are empty because fields are hidden)
- Supported versions in the readme is now limited to 11,12,13. We will most likely help out with the older versions, but spanning 5 versions of active support is a bit too much
- laravel 6 is not supported in version 13
- The Audit may no longer ONLY be for eloquent events, but an AuditCustom event may be dispatched. This provides a way to solve many of the cases where eloquent is missing events. For example when using the attach,detach and sync methods to work with pivots.

There ARE issues to note when upgrading
1. config key "resolver" is gone and is instead called "resolvers"
2. the user resolver has moved to the user config array
3. Resolvers now implements the resolver contract instead of a specific interface for each resolver


anteriovieira:
Looks good to me!

PaolaRuby:
Maybe here drop php 7 and start using php 8 sintaxis

MortenDHansen:
> Maybe here you can drop php 7 and start using php 8 syntax PHP 7.4 had active support until November 28, 2021, now only security support

I would love to, but i feel it's a bit too early to cut Laravel 7 support (php ^7.2.5).

fredsal:
>>you can drop php 7 and start using php 8 syntax

>i feel it's a bit too early to cut Laravel 7 support (php ^7.2.5).

Laravel 7.x had active support until  March 3rd, 2021, also v12 still has active support for 6.x,7.x, mayor releases are for letting behind discontinued versions and bring breaking changes

image

MortenDHansen:
> > > you can drop php 7 and start using php 8 syntax
> 
> > i feel it's a bit too early to cut Laravel 7 support (php ^7.2.5).
> 
> Laravel 7.x had active support until March 3rd, 2021, also v12 still has active support for 6.x,7.x, mayor releases are for letting behind discontinued versions and bring breaking changes
> 
> image
> 
> Laravel 7.x supports php 8 sintax https://github.com/laravel/framework/blob/88dd075b9f1beaff93796a0ee51818796ca04654/composer.json#L18

Yeah, but laravel 7 and 8 is compatible with php 7.3, and by extension we should be as well https://github.com/laravel/framework/blob/8.x/composer.json 

I don't mind though. There is plenty to update before we need to cut 7.3 support 😃

Issue #568, opened 2020-01-24
Missing casting on old_values


orim181:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 5.7.25
| Package version   | 8.0.4
| PHP version       | 7.2.14

#### Actual Behaviour

The old values aren't casted but the new values are:
![image](https://user-images.githubusercontent.com/11587953/73098488-83d60500-3eb7-11ea-8976-d9e19da88b20.png)

#### Expected Behaviour

I expected a consistent behavior so, for example, having the old_values to be casted like the new_values.

#### Steps to Reproduce

1. In the eloquent model that implements Auditable, leverage Laravel's casting feature:
```
protected $casts = [
        'is_active' => 'boolean'
]
```
2. Simulate a change on that field and look at the audit table. You'll see inconsistent casting

#### Possible Solutions

I leveraged the `transformAudit` feature of the library and created the following trait to apply the casting on the old_values:
```
trait AuditRecordCastFix
{
    public function transformAudit(array $data): array {
        foreach ($data['old_values'] as $key => $value) {
            if($this->hasCast($key)) {
                $data['old_values'][$key] = $this->castAttribute($key, $value);
            }
        }
        return $data;
    }

}
```

..then referenced it on my eloquent models:
```
class Blah extends Model implements Auditable {
   // use \OwenIt\Auditing\Auditable;
   use \OwenIt\Auditing\Auditable, Traits\AuditRecordCastFix {
        Traits\AuditRecordCastFix::transformAudit insteadof \OwenIt\Auditing\Auditable;
    }
}
```

orim181:
When I troubleshooted this, the issue is that the auditing library is populating OLD and NEW via this method in its `src\Auditable` class...
![image](https://user-images.githubusercontent.com/11587953/73100142-f85e7300-3eba-11ea-87c6-b64a2caf519b.png)

...but Laravel only casts `attributes` not the `original` properties.

vitomeuli:
Similar problem for me, with 'date' fields.
My stack is much younger than the OP's:
- laravel/framework v7.5.2
- owen-it/laravel-auditing v10.0.0
- PHP 7.4.4

christianbayer:
This occurs because the method is accessing the $original property directly. If it was using the getOriginal() method instead, the cast would be working.

MortenDHansen:
[...]

Issue #216, opened 2017-03-20
0 and False for boolean marked as a change


jorenvh1:
Is there a way to change the configuration for validating booleans?
I send False to my DB but it will be stored as 0. Because of this behaviour, it will be marked as a change when it's not.

Thanks in advance

quetzyg:
I'll look into it.

jorenvh1:
I currently fixed it by just replacing 'False' with 0 and 'True' with 1 so it doesn't get logged as a change.
Still interested in how you'd fix this problem though ;)

quetzyg:
Hey @jorenvh1,

This is an issue with Eloquent's `getDirty()` method. I guess we'll just have to wait until it gets a proper fix. 
Meanwhile, you'll have to resort to the workaround you described.

If you're interested, you can also track how things are progressing in the current [open issue](https://github.com/laravel/internals/issues/349) about this problem, which has links [to](https://github.com/laravel/framework/issues/16823) [other](https://github.com/laravel/framework/issues/15060) [closed](https://github.com/laravel/framework/issues/10735) [issues](https://github.com/laravel/framework/issues/8972).

Marking this as closed.

LeonAlvarez:
@quetzyg according to https://github.com/laravel/ideas/issues/349 it should work, but I still facing it on v9 laravel 5.8

quetzyg:
@LeonAlvarez, that has been fixed since version 5.5. If you're still having a similar issue, you're either not on the versions you think you are, or something else is wrong with your code.

See the [troubleshooting](http://www.laravel-auditing.com/docs/9.0/troubleshooting) for a workaround, using [this](https://gist.github.com/crashkonijn/7d581e55770d2379494067d8b0ce0f6d) trait.

**PS:** I'm no longer a collaborator to this package, since I started working on my own auditing package.

LeonAlvarez:
Sorry, I found the issue had a typo in the property add to $casts, ty

PR #688, opened 2022-02-17
fix: diff attributes casting

christianbayer:
Solves #568, related to #216.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -153,8 +153,8 @@ trait Auditable
 
         foreach ($this->getDirty() as $attribute => $value) {
             if ($this->isAttributeAuditable($attribute)) {
-                $old[$attribute] = Arr::get($this->original, $attribute);
-                $new[$attribute] = Arr::get($this->attributes, $attribute);
+                $old[$attribute] = $this->getOriginal($attribute);
+                $new[$attribute] = $this->getAttribute($attribute);
             }
         }
 


MortenDHansen:
I don't actually think this is still an issue. 
The following test seems to cover the case

```
    /**
     * @test
     * @return void
     */
    public function itHandlesJsonColumnsCorrectly()
    {
        $article = factory(Article::class)->create(['config' => ['articleIsGood' => true, 'authorsJob' => 'vampire']]);
        $article->refresh();

        $article->config = ['articleIsGood' => false, 'authorsJob' => 'vampire'];
        $article->save();

        /** @var Audit $audit */
        $audit = $article->audits()->skip(1)->first();

        $this->assertSame(false, $audit->getModified()['config']['new']['articleIsGood']);
        $this->assertSame(true, $audit->getModified()['config']['old']['articleIsGood']);
    }
```

MortenDHansen:
I think the issue is resolved 

The following test should cover it.
```
    public function itHandlesJsonColumnsCorrectly()
    {
        $article = factory(Article::class)->create(['config' => ['articleIsGood' => true, 'authorsJob' => 'vampire']]);
        $article->refresh();

        $article->config = ['articleIsGood' => false, 'authorsJob' => 'vampire'];
        $article->save();

        /** @var Audit $audit */
        $audit = $article->audits()->skip(1)->first();

        $this->assertSame(false, $audit->getModified()['config']['new']['articleIsGood']);
        $this->assertSame(true, $audit->getModified()['config']['old']['articleIsGood']);
    }
```

I also tried executing this:
```
    /**
     * @test
     * @return void
     */
    public function itCastsJson()
    {
        $article = factory(Article::class)->create(['reviewed' => false]);

        $article->reviewed = true;
        $article->save();

        $audit = $article->audits()->get()->last();
        $this->assertEquals('{"reviewed":false}', json_encode($audit->old_values));
        $this->assertEquals('{"reviewed":true}', json_encode($audit->new_values));
    }
```

christianbayer:
This problem still exists. Maybe the test case is not covering all possibilities. Without the patch I'm proposing in this PR, the cast is not being applied. See my comment at https://github.com/owen-it/laravel-auditing/issues/568#issuecomment-1042514492.

MortenDHansen:
> This problem still exists. Maybe the test case is not covering all possibilities. Without the patch I'm proposing in this PR, the cast is not being applied. See my comment at [#568 (comment)](https://github.com/owen-it/laravel-auditing/issues/568#issuecomment-1042514492).

Any chance you can help me with a scenario where it fails so that i can cover it with a test 😄

MortenDHansen:
Just to be clear, problem in the issue is that the old values and new values are different, which i can't reproduce, but i may be missing something. What we expect in the columns in the database is the raw representation. In this case 0 or 1 (boolean in mysql). When fetching the record using ->getModified() we would expect that to run through the cast. The values stored in the audits table are not supposed to be cast before they are stored.

christianbayer:
I'll try to explain my scenario. Just for a background, we have a Laravel 8 application with MariaDB 10.3 and Auditing v12.0. For example, we have a `users` table:
```
Schema::table('users', function (Blueprint $table) {
    $table->bigIncrements('id');
    $table->string('name', 255);
    $table->boolean('active')->default(1);
});
```
And a user model:
```
class User implements AuditableContract
{
    use Auditable;
    protected $fillable = [
        'name',
        'active',
    ];
    protected $casts = [
        'active' => 'boolean',
    ];
}
```
An important detail: the `boolean` defined in the migration does not really creates a boolean column in the database. Instead it is a tinyint column. I don't know why, but that's the way Laravel handles it.

Anyway. If I create a User called Ana with active="true", the Audit will store: `old` as `[]` and `new` as `{name: "Ana", active: "true"}`. Now, I will just open the form, click save (without changing nothing, still Ana, still active), and the Audit will pick a change of `old` as `{active: 1}` and `new` as `{active: true}`.

Why? **Because `$this->original` does not cast any attributes, but the `$this->getOriginal()` does.** 

That's my use case, that's my issue. After changing what I've proposed in this PR, this issue has gone away.

If you need more info, or anything else I can help with, please let me know.

MortenDHansen:
Allright. Thanks.

There is no boolean in mysql (mariadb). It's just a tinyint field. The underlying value will be 0,1 or null. 

The laravel-auditing package will take the user supplied values in changes (before write) and the values from the model (fetched) out of the Eloquent model, but based on the getDirty(). In Laravel, that uses Laravels "originalsEquivalent()" method analyzing the casts (and various cases). What is written by audit is determined by the compared cast values from eloquent, but the audit content is the users values.

Basically, laravel auditing will work "as intended", but using the users input for what is supplied against the existing record. I see that it is a problem when the existing column is a boolean it can be written arbitrarily into the serialized json, but i can't find a problem where a wrong update will occur. I've tried several approaches. In any case, that would be an Eloquent problem since the update is triggered by Eloquent, not laravel-audit.

```
    /**
     * @test
     * @return void
     */
    public function itCastsJson()
    {
        $article = factory(Article::class)->create(['reviewed' => true]);

        // Set true to true
        $article->reviewed = true;
        $article->save();

        // set true to string true (evaluates to true)
        $article->reviewed = "true";
        $article->save();

        // Set to database equivalent of true (1)
        $article->reviewed = 1;
        $article->save();

        // Set to string false (evaluates to true)
        $article->reviewed = "false";
        $article->save();

        // Check that none of the 4 previous saves spawned an audit record (nothing actually changed)
        $audits = $article->audits()->get();
        $this->assertCount(1, $audits);

        // Assert the audit new value is true as created
        $audit = $article->audits()->get()->last();
        $this->assertEquals(["new" => true], $audit->getModified()['reviewed']);
        $article->refresh(); // reset experiments

        // Change the value of the field
        $article->reviewed = false;
        $article->save();

        $audit = $article->audits()->get()->last();
        $this->assertEquals('updated', $audit->event);

        // Values of getModified are correct booleans
        $this->assertEquals(['new' => false, 'old' => true], $audit->getModified()['reviewed']);


        // Change the value of the field, but with a different type
        $article->reviewed = 'banana'; // This is a bad way to write a boolean value since any string will be true
        $article->save();

        // assert new value is user input value, and old is still boolean type.
        $audit = $article->audits()->get()->last();
        $this->assertEquals(["new" => true, "old" => false], $audit->getModified()['reviewed']);
    }
```

Article::class casts reviewed to boolean

In terms of the PR, it would be a breaking change so i can't merge it as is. The system should interpret the values based on casts when fetching the content. Not on write. I see that the json serializer however may act in unpredictable ways when storing the json, but I am still not able to reproduce the problem.

aromka:
Running into a similar issue where it creates audits for changes from `{ value: 0 }` to `{ value: false }`

Screenshot_430

Is there a method where these values can be compared and removed the audits?

parallels999:
@aromka  try casting
```php
protected $casts = ['value'=>'integer'];
```
Also try `transformAudit`
https://laravel-auditing.com/guide/audit-transformation.html

PR #692, opened 2022-02-22
Dont publish migration if migration exists

MortenDHansen:
As suggested with pr #644

### DIFF
--- a/src/AuditingServiceProvider.php
+++ b/src/AuditingServiceProvider.php
@@ -4,10 +4,9 @@ namespace OwenIt\Auditing;
 
 use Illuminate\Contracts\Support\DeferrableProvider;
 use Illuminate\Support\ServiceProvider;
-
-use OwenIt\Auditing\Contracts\Auditor;
 use OwenIt\Auditing\Console\AuditDriverCommand;
 use OwenIt\Auditing\Console\InstallCommand;
+use OwenIt\Auditing\Contracts\Auditor;
 
 class AuditingServiceProvider extends ServiceProvider implements DeferrableProvider
 {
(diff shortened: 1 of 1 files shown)


src/AuditingServiceProvider.php
```diff
@@ -52,11 +51,13 @@ private function registerPublishing()
                 __DIR__ . '/../config/audit.php' => base_path('config/audit.php'),
             ], 'config');
 
-            $this->publishes([
-                __DIR__ . '/../database/migrations/audits.stub' => database_path(
-                    sprintf('migrations/%s_create_audits_table.php', date('Y_m_d_His'))
-                ),
-            ], 'migrations');
+            if (!class_exists('CreateAuditsTable')) {
```
anteriovieira, line 54: check if the file exists instead of the class.

anteriovieira (changes requested):

MortenDHansen:
> 

Yeah, but i don't think you actually need to use the anonymous migrations if you don't want to. If we expect all existing migrations to be created as a class and we dont change how we publish it, then the class check should be enough. Spatie seems use same approach: https://github.com/spatie/laravel-translation-loader/blob/main/src/TranslationServiceProvider.php

anteriovieira:
Looks good to me! thanks

sarukomine:
@MortenDHansen the change is not working for me, at least not working for Laravel Framework 9.9.0, I still can keep on publishing migration files. Maybe we should change the way to check if file exists instead of the class.

MortenDHansen:
Yeah, maybe my brilliant simple fix was not as brilliant as i figured, and a bit too simple :/

PR #695, opened 2022-03-03
Default user guard

MortenDHansen:
Laravel Auth changed and it would be nice to change default guard. New laravel installs throws an exception from the previous default api guard, but to avoid breaking backwards we could skip unavailable guards and set a sensible default


falko189:
Thanks for looking into this! Few style notices to fix!

Issue #698, opened 2022-03-11
Is it ok to have `$auditEvent` protected when using customAudit?


r0lodex:
https://github.com/owen-it/laravel-auditing/blob/1ec0aabc97d660292679ff1045a1369b11c68912/src/Auditable.php#L34

I've come across the error `Unable to handle "" event, () method missing` when trying to use `AuditCustom`. I think it should be a public property.

r0lodex:
#699

MortenDHansen:
You are totally right! - It looks like a typing mistake 😞  

Good catch! - thanks!

PR #699, opened 2022-03-11
`$auditEvent` from protected to public to make AuditCustom possible

r0lodex:
As per this issue — https://github.com/owen-it/laravel-auditing/issues/698

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -31,7 +31,7 @@ trait Auditable
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


MortenDHansen:
Oh! 
Good call! - How did i miss that 😕 - Let me just check if we could run into other issues...

MortenDHansen (approved):

MortenDHansen:
Closes issue #698

PR #706, opened 2022-03-28
Don't audit if there aren't any changes (except for retrieving models)

madman-81:
This PR prevents auditSync() to create a new record when there are no changes. 

Before, if a call was made to auditSync(), it would create a new audit record, regardless if there where any changes or not. 

This query shouldn't return any records:
``` sql
select *
from audits 
where event = 'sync'
and old_values = new_values
```

### DIFF
--- a/src/Auditor.php
+++ b/src/Auditor.php
@@ -78,6 +78,14 @@ class Auditor extends Manager implements Contracts\Auditor
             }
         }
 
+        //Don't store audits if there are no changes, expect for 'retrieved' events
+        if (
+            $model->toAudit()['new_values'] == $model->toAudit()['old_values'] &&
+            !in_array($model->getAuditEvent(), ['retrieved'])
+        ) {
+            return;
+        }
+
         $audit = $driver->audit($model);
         if (!$audit) {
             return;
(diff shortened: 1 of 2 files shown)


mitchdesign:
We were literally today working on a similar issue. We made an addition in the 'auditSync' logic to be able to jump out if there are no changes : https://github.com/owen-it/laravel-auditing/pull/707

The (dis)advantage of that is that it will not effect other types of events (including custom ones) which might not have changes to any attributes but still should be audited.

I saw your tests were doing what we want so I copied them, I hope that is ok.

madman-81:
> We were literally today working on a similar issue. We made an addition in the 'auditSync' logic to be able to jump out if there are no changes : #707
> 
> The (dis)advantage of that is that it will not effect other types of events (including custom ones) which might not have changes to any attributes but still should be audited.
> 
> I saw your tests were doing what we want so I copied them, I hope that is ok.

Well, at least is shows that it's a issue more people face and would like to get fixed. Let's see what which one the maintainers prefer.

Of course it's fine to use the tests 👍

MortenDHansen:
Hi, thanks for the PR! - Good solution! 😄 

I merged the other one, but i'm leaving this one open for a bit to remember it since i am still messing about with a refactor.

MortenDHansen:
Hi, thanks for the PR! - Good solution! 😄 

I merged the other one, but i'm leaving this one open for a bit to remember it since i am still messing about with a refactor.

PR #707, opened 2022-03-28
Add option to skip auditSync when sync has no changes

mitchdesign:
We found that often we are storing models to make minor changes but some relationships are not changed. If we do a `auditSync` call, we will get an audit even if the relationship was not affected. 

This PR adds a method `auditSyncIfChanged` which will abort the audit when it finds the relationship sync() call had 0 attached, updated or detached items.

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -96,13 +96,18 @@ return [
     |
     | Some events may be empty on purpose. Use allowed_empty_values to exclude
     | those from the empty values check. For example when auditing
-    | model retrieved events which will never have new and old values
+    | model retrieved events which will never have new and old values.
+    |
+    | When using empty_values => true, you can still exclude certain events
+    | with empty values by specifying them in disallowed_empty_values.
     |
     */
 
     'empty_values'         => true,
     'allowed_empty_values' => [
         'retrieved'
+    ],
+    'disallowed_empty_values' => [
     ],
 
     /*
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -681,6 +681,8 @@ trait Auditable
             throw new AuditingException('Relationship ' . $relationName . ' was not found or does not support method sync');
         }
 
+        $this->auditEvent = 'sync';
+
         $this->auditCustomOld = [
             $relationName => $this->{$relationName}()->get()->isEmpty() ? [] : $this->{$relationName}()->get()->toArray()
         ];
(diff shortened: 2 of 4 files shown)


MortenDHansen:
Its actually a big miss on my part. That should probably be the default behaviour. A sync with no changes should not be audited. I'm thinking about if there should even be an explicit method 🤔

mitchdesign:
Thanks.

To me it would be reasonable default behavior to not audit when no change happens in the sync.  But I did not want to break existing behavior, that is why I added it like this.

If it is made automatic then there could be a setting similar to `audit.empty_values`, to do it or not. Like `audit.empty_sync`.

mitchdesign:
I updated the proposal to be (sort of) a mix of the #706 from @madman-81 and mine.

The idea is:
- In a 'auditSync' call, we check if the sync produced any changes
- If no changes, then we set the old and new values to be `[]` instead of being twice the same list of child models
- This way, the `Auditor->execute()` will be able to handle the sync like any other event and check for empty values

I added a config option to make it possible to store empty values, but still exclude certain events from storing when empty. 

I think this flow matches better with the conceptual design. An event will happen but can be skipped from auditing just like is already possible. Which I think is cleaner than having separate logic inside the auditSync.

mitchdesign:
@MortenDHansen How do you feel about the last implementation? This means sending empty oldvalues and newvalues, if a sync did not lead to changes. And then we let settings decide if empty audits should be recorded or not.

MortenDHansen:
> @MortenDHansen How do you feel about the last implementation? This means sending empty oldvalues and newvalues, if a sync did not lead to changes. And then we let settings decide if empty audits should be recorded or not.

Thank you for making an implementation! - I am sorry for slow responses. We have a madhouse of sick kids around here. I want to pull it down to test it, and hope to get around to it during the day. I do wonder if there is a way around adding the config value so i may make a few adjustments.

mitchdesign:
@MortenDHansen Not at all, we all get sick (and with kids kids even more so).

Indeed the config part might not be needed, good defaults could be enough if it is decided that "a sync with no changes just isn't an auditable event". 

If you have an idea but no time to tinker, feel free to respond and I can make changes too :)

MortenDHansen (approved):

madman-81:
@MortenDHansen When can we expect a new release with this update? I'm looking forward to using it in my project.

PR #710, opened 2022-04-06
Handle default timezones properly (fixes #709)

### LINKED ISSUE #709: When using a default timezone other than UTC and calling transitionTo, datetime values are shifted by the timezone modifier
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 8.83.7
| Package version   | 13.0.2
| PHP version       | 7.4.16

#### Actual Behaviour

I have the default timezone for my project set to `"America/New_York"` in the `app.timezone` config option. Prior to [Laravel changing the way date serialization works in 7.x](https://laravel.com/docs/7.x/upgrade#date-serialization), the package was functioning normally.

Now if I transition to an audit and a date value on the model is loaded from the audit, the date value is shifted forward in time by 5 hours (the inverse of the timezone offset).

I originally mentioned this issue in the Discord and explained it in much greater detail - you can find that detailed explanation in [this pastebin.](https://pastebin.com/BUSGeGS3)

The quick version of the explanation is that the `getModified` function on the `Audit` class treats any value that implements `DateTimeInterface` differently from other values. Specifically if a value implements `DateTimeInterface` it will be passed through the auditable's `serializeDate` function instead of simply being returned like other values. With the new Laravel behavior as of 7.x, `serializeDate` converts the timezone to UTC and returns an ISO 8601 datetime string which ends in Z to indicate that it's UTC. When this value is eventually passed to the `setAttribute` function the timezone is ignored and the default timezone is applied, because that function expects any dates to be passed in using the datetime format for the current database.

#### Expected Behaviour

Date values should be transitioned to the value stored on the audit and maintain their original value, and ideally their original timezone.

#### Steps to Reproduce

I [posted a repository to Github](https://github.com/tmas/laravel-audit-timezone-bug) that reproduces the issue, including a test showing exactly what happens:
1. `git clone git@github.com:tmas/laravel-audit-timezone-bug`
2. `cd laravel-audit-timezone-bug`
3. `composer install`
4. Set up a database connection
5. `php artisan migrate`
6. `php artisan test`

I have changed 2 of the configuration options in that project. First, `app.timezone` is set to `"America/New_York"`. This is required to reproduce the issue. Second, `audit.console` is set to `true`. This is not required to reproduce the issue, but makes it much easy to poke around using `php artisan tinker`.

I'll also link the pastebin I mentioned in a previous section again since it seems like it fits better here: https://pastebin.com/BUSGeGS3

#### Possible Solutions
I propose removing this section from the `getModified` function:
```
if ($value instanceof DateTimeInterface) {
    $modified[$attribute][$state] = !is_null($this->auditable)
        ? $this->auditable->serializeDate($value)
        : $this->serializeDate($value);
}
```

In my testing, everything works properly if that code is removed. I'm not sure what made it necessary in the first place, but passing a value that implements `DateTimeInterface` to the `setAttribute` function works fine and this code block converts the `DateTimeInterface` to a string which isn't handled as expected when passed to `setAttribute`. As far as I can tell the `getModified` function is only ever called in one place in the `transitionTo` function on the `Auditable` trait, so I'm fairly certain removing this code block won't have any side effects.

tmas:
Return values implementing `DateTimeInterface` directly in `getModified` instead of serializing before passing to the `Auditable`'s `setAttribute` function.

`setAttribute` handles values that implement `DateTimeInterface` properly, but does not handle serialized dates properly. Since Laravel 7 dates are serialized to a format like:
```"2022-04-06T12:53:18Z"```

`setAttribute` expects dates to be in the same format the database would use, which is typically something like:
```"2022-04-06 12:53:18"```

Because the expected format does not include a timezone, the timezone is ignored. When a default timezone is set, that means that the datetime value gets converted to UTC by the `serializeDate` function, then gets unserialized and treated as the default timezone by `setAttribute`. Ultimately, `"2022-04-06 12:00:00 America/New_York"` becomes `"2022-04-06T17:00:00Z"` becomes `"2022-04-06 17:00:0 America/New_York"`.

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -243,10 +243,6 @@ trait Audit
 
             $value = $this->getDataValue($key);
             $modified[$attribute][$state] = $value;
-
-            if ($value instanceof DateTimeInterface) {
-                $modified[$attribute][$state] = !is_null($this->auditable) ? $this->auditable->serializeDate($value) : $this->serializeDate($value);
-            }
         }
 
         return $json ? json_encode($modified, $options, $depth) : $modified;


MortenDHansen:
Thank you for contributing! 

Looks like a good fix! - I need a test proofing that this works as intended, and then i need to go over the failing test to ensure it doesn't break something users rely on. You are very welcome to help with a test, else i will get around to it as soon as possible. 🙂

erikn69:
was it already added?

https://github.com/owen-it/laravel-auditing/commit/82b1cdd814b3fb0f5a47a35ea929a34776e80238
https://github.com/owen-it/laravel-auditing/blob/240b29a1e6c40c02ed1fb9b4c01ef94d6d4ff7c0/src/Audit.php#L247-L249

MortenDHansen:
@erikn69 im not sure. I do remember fiddling with the date serialization, but i cant from looking at this identify if its a solved problem or something that needs checking. The date serialization is a bit annoying because i want to keep it working "like it used to" so that we dont break expectation. I haven't had the opportunity to check it :)

parallels999:
Already fixed on #775

PR #719, opened 2022-06-24
User morph prefix from config on migration

erikn69:
On migration use setted morph prefix from config, same as other places
https://github.com/owen-it/laravel-auditing/blob/2317a5ed72f0c4256b1c274d3e6ef53db5d0598c/src/Auditable.php#L320
https://github.com/owen-it/laravel-auditing/blob/2317a5ed72f0c4256b1c274d3e6ef53db5d0598c/src/Auditable.php#L332-L333

### DIFF
--- a/database/migrations/audits.stub
+++ b/database/migrations/audits.stub
@@ -2,6 +2,7 @@
 
 use Illuminate\Database\Migrations\Migration;
 use Illuminate\Database\Schema\Blueprint;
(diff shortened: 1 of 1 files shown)


angeljqv:
@anteriovieira @MortenDHansen ping, looks good to me

MortenDHansen (approved):
Yes. That is how it should be :)

PR #733, opened 2022-07-29
Fix old resolvers config deprecation

erikn69:
Deprecation not means break old functionality

https://en.wikipedia.org/wiki/Deprecation
>Typically, deprecated materials are not completely removed to ensure legacy compatibility or back up practice ...
>It can also imply that a feature, design, or practice will be removed or discontinued entirely in the future.

Same logic from `UserResolver`: (it get the value from old deprected config)
https://github.com/owen-it/laravel-auditing/blob/240b29a1e6c40c02ed1fb9b4c01ef94d6d4ff7c0/src/Auditable.php#L360-L366

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -375,13 +375,14 @@ trait Auditable
     protected function runResolvers(): array
     {
         $resolved = [];
-        if (Config::has('audit.resolver')) {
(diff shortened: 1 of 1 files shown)


MortenDHansen (approved):

MortenDHansen:
Right!

PR #740, opened 2022-09-09
Removed User from comment

KuenzelIT:
Nothing important, just cleaning up the comment


parallels999 (commented):
@anteriovieira this does not change the functionality, it is only informative

anteriovieira:
thanks guys!

Issue #588, opened 2020-05-27
Value Object Casting issue. + Recommendation.


rrpadilla:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | yes
| Framework         | Laravel|Lumen
| Framework version | 7.13.0
| Package version   | 10.0.0
| PHP version       | 7.4.5

#### Actual Behaviour

Following your documentation: "The **getModified()** method honors all attribute mutators and/or casts defined in the Auditable model, transforming the data accordingly."
The issue is when casting to **value objects**. See [laravel documentation](https://laravel.com/docs/7.x/eloquent-mutators#custom-casts).
When this method is calling "**[getFormattedValue](https://github.com/owen-it/laravel-auditing/blob/master/src/Audit.php#L118-L136)**" laravel will get the value from the **classCastCache** property. See: https://github.com/laravel/framework/blob/master/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L562

**$modified['old']** will be equal to **$modified['new']** if your attribute is using a custom cast.
```php
@foreach ($audit->getModified() as $attribute => $modified)
        <ul>
            <li>OLD: {{ $modified['old'] }}</li>
            <li>NEW: {{ $modified['new'] }}</li>
        </ul>
@endforeach
```

#### Expected Behaviour

**$modified['old']** should NOT be equal to **$modified['new']** if the attribute is using a custom cast.

#### Steps to Reproduce

1- First create a cast class
```php
<?php
namespace App\Casts;

use App\Address;
use Illuminate\Contracts\Database\Eloquent\CastsAttributes;

class Address implements CastsAttributes
{
    /**
     * Cast the given value.
     *
     * @param  \Illuminate\Database\Eloquent\Model  $model
     * @param  string  $key
     * @param  mixed  $value
     * @param  array  $attributes
     * @return \App\Address
     */
    public function get($model, $key, $value, $attributes)
    {
        return new Address(
            $attributes['address_line_one'],
            $attributes['address_line_two']
        );
    }

    /**
     * Prepare the given value for storage.
     *
     * @param  \Illuminate\Database\Eloquent\Model  $model
     * @param  string  $key
     * @param  \App\Address  $value
     * @param  array  $attributes
     * @return array
     */
    public function set($model, $key, $value, $attributes)
    {
        return [
            'address_line_one' => $value->lineOne,
            'address_line_two' => $value->lineTwo,
        ];
    }
}
```

[...]

PR #742, opened 2022-09-15
Unset classCastCache for given key

bskl:
"[getFormattedValue](https://github.com/owen-it/laravel-auditing/blob/master/src/Audit.php#L122-L148)" method is calling laravel will get the value from the classCastCache property. So old and new values are the same. This PR unset the classCastCache for given key so old value is not retrieve from cache.
Fix issue: #588

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -141,11 +141,11 @@ trait Audit
 
         // Cast to native PHP type
         if ($model->hasCast($key)) {
-            unset($model->classCastCache[$key]);
-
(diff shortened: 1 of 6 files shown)


parallels999:
Hi, could you add some tests?

bskl (commented):
Add test.

erikn69:
I don't have much experience with custom cast, but I went through the laravel code and couldn't find any problem.
It does not change the functionality for those who do not use that feature

MortenDHansen (approved):

PR #743, opened 2022-09-28
Update IpAddressResolver.php

lincolnpjames:
if x-forwarded-for header is present return it instead of ip which would generally be 127.0.0.1

### DIFF
--- a/src/Resolvers/IpAddressResolver.php
+++ b/src/Resolvers/IpAddressResolver.php
@@ -10,6 +10,6 @@ class IpAddressResolver implements Resolver
 {
     public static function resolve(Auditable $auditable): string
     {
-        return Request::ip();
+        return Request::header('x-forwarded-for', Request::ip());
     }
 }


parallels999 (approved):

parallels999:
@anteriovieira it is a good fix

anteriovieira:
thanks guys!

ceejayoz:
This change breaks applications with multiple proxies.

`Request::ip()` already uses `X-Forwarded-For` automatically. @lincolnpjames should resolve their issue by properly configuring their `App\Http\Middleware\TrustProxies` middleware to accept the header from their proxy.

APavlov2:
Reverting this change caused the package to break the recording of all activities. The Reqeust::ip() returns the IP address formatted as 127.0.0.1:44444 format (adding the port number at the end) while the migration that generates the audit table expects a flat IP format. 
This throws a fatal error when performing the insert operation because of the SQL validation. 

A workaround is having a custom IpResolver, but it would be good if the package handles the issue as other systems already deployed might be impacted by this change on the next update/re-deploy.

ceejayoz:
@APavlov2 The package has been using `Request::ip()` since five years ago. https://github.com/lincolnpjames/laravel-auditing/blame/master/src/Resolvers/IpAddressResolver.php

Only the most recent release used the different methodology from this PR.

lincolnpjames:
I see the issue and understand, I will explore the custom resolver as
suggested.

Also curious how the port came with the IP? Assuming that it reading the
x-forwarded-for header, the Mozilla docs suggest that only IP’s should be
contained in this however, it also raises another  issue that multiple
values can be supplied. More at
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

Another thing to cross my mind is that is there is an issue storing the
below example, does that mean IPv6 is not supported. When I get some time I
might explore this to find out.

ceejayoz:
@lincolnpjames Before you go down the custom resolver route, try the trusted proxy route. It's safer from a security standpoint, and requires no custom code if configured correctly.

ceejayoz:
@APavlov2 I did some checking, and I think something's odd on your end. A new Laravel install using `php artisan serve` to host on `http://localhost:8000/` still shows `127.0.0.1` as the `Request::ip()` value. Looking at the raw `$_SERVER`, I see a `REMOTE_ADDR` of `127.0.0.1` and a `SERVER_PORT` of `8000`.

I suspect something odd in your server configuration. Check your Apache/Nginx config to see if you're manipulating the `REMOTE_ADDR` header anywhere.

parallels999:
yes, seems like there is something on @APavlov2 environment

PR #749, opened 2022-10-30
Added $auditTags for flexible adding of tags

mvdgeijn:
Adding this variable and returning it from generateTags makes is much more flexible to set tags from the model class.


anteriovieira:
Closing this for now! We can check it again later.

PR #758, opened 2023-01-18
Fix disable auditing before model initialization

evgeek:
The `Auditable::bootAuditable()` trait loader on the model works only once, at the very first initialization of the class. Thus, at the moment, if auditing is disabled on the model during initialization, an observer will not be booted on the model, and a further call to `Model::enableAudit()` will not change anything - there is no observer, no events are registered.

The `itDisablesAndEnablesAuditingBackAgain` test skipped this behavior because the model was first initialized before auditing was disabled.

I suggest init observer on the model, regardless of the `$auditingDisabled` flag. I made the appropriate changes, and the `itDisablesAndEnablesAuditingBackAgainViaFacade` test, where auditing is disabled before initialization.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -63,7 +63,7 @@ trait Auditable
      */
     public static function bootAuditable()
     {
-        if (!self::$auditingDisabled && static::isAuditingEnabled()) {
+        if (static::isAuditingEnabled()) {
             static::observe(new AuditableObserver());
         }
     }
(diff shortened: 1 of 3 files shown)


anteriovieira:
@parallels999 @erikn69 @MortenDHansen any comments here?

erikn69:
Makes a lot of sense to me, this would only happen in this specific case, it makes me think that no one has ever used `disableAuditing()` in that way before.

So `bootAuditable` from `Model A` class is called once after `disableAuditing()` and before `enableAuditing()` the same `Model A` stops auditing because there is no observer

PR #762, opened 2023-01-28
Feat/add async support

hamedmax73:
You may set a default exclude set in the config.

1- first config this setups:

```
'should_queue' => true,

'job_queue_name' => 'audit',
```

2- you can disable this setup in your model

This is done on a per `Auditable` model basis, by assigning a `bool` value to the `$shouldQueue` attribute.

```
    protected bool $shouldQueue = false;
```

3- after that you must run your queue

----
in case of manually publish entity changes use this sample:
```
        User::disableAuditing();
            //user mmodel update or create or ...
        User::enableAuditing();
            //save it manually
        if ($audit = Auditor::execute($createdUser)) {
            Auditor::prune($createdUser);
        }

```

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -88,7 +88,7 @@ trait Auditable
     {
         $this->excludedAttributes = $this->getAuditExclude();
 
-        // When in strict mode, hidden and non visible attributes are excluded
+        // When in strict mode, hidden and non-visible attributes are excluded
         if ($this->getAuditStrict()) {
             // Hidden attributes
             $this->excludedAttributes = array_merge($this->excludedAttributes, $this->hidden);
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -545,6 +545,15 @@ trait Auditable
     }
 
     /**
+     * check we should queue the audit or not
+     * @return bool
+     */
+    public function getShouldQueue(): bool
+    {
+        return $this->shouldQueue ?? Config::get('audit.should_queue', true);
+    }
+
+    /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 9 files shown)


src/Auditable.php
```diff
@@ -329,15 +342,15 @@ public function toAudit(): array
...
-            $morphPrefix . '_id'   => $user ? $user->getAuthIdentifier() : null,
-            $morphPrefix . '_type' => $user ? $user->getMorphClass() : null,
-            'tags'                 => empty($tags) ? null : $tags,
-        ], $this->runResolvers()));
+            'old_values'         => $old,
+            'new_values'         => $new,
+            'event'              => $this->auditEvent,
+            'auditable_id'       => $this->getKey(),
+            'auditable_type'     => $this->getMorphClass(),
+            $morphPrefix.'_id'   => $user ? $user->getAuthIdentifier() : null,
```
parallels999, line 350: Why so many spaces removed? I find it unnecessary to change the writing style, it makes this change hard to read

MortenDHansen:
Thank you for the PR. First up, it fails all tests because the changes depends on resolving 'events'. 

In order to introduce this, we are going to need some sort of test coverage. I am not certain it will actually work in async as it is right now, it might, but i am not certain. 

I am also curious as to what caveats there might be using the queues. For example, some queue implementations have limits to the message size (eg. sqs KiB kib), and not all. models may fit into that. 

I would also like to investigate if the performance impact by audit is big enough to warrant this effort, and which part of the process is the culprit.

As it is for now, this is not ready to ship. - but thank you for the effort to create an example.

PR #774, opened 2023-03-09
Add DumpResolver for empty values when is needed

angeljqv:
https://github.com/owen-it/laravel-auditing/blob/75b33ec9b4345e4539d94d6c28233b34c5d2730f/config/audit.php#L44-L48
If we don't want any data
```php
'resolvers' => [ 
     'ip_address' => OwenIt\Auditing\Resolvers\DumpResolver::class, 
     'user_agent' => OwenIt\Auditing\Resolvers\DumpResolver::class, 
     'url'        => OwenIt\Auditing\Resolvers\DumpResolver::class, 
 ], 
```

### DIFF
--- a/src/Resolvers/DumpResolver.php
+++ b/src/Resolvers/DumpResolver.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace OwenIt\Auditing\Resolvers;
+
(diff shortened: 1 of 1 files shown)


anteriovieira:
thanks @angeljqv

PR #798, opened 2023-03-14
fixing phpstan namespace issues

aglipanci:
``` 
Class OwenIt\Auditing\Contracts\OwenIt\Auditing\Models\Audit was not found while trying to analyse it - discovering symbols is probably not configured properly.  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols                  
```

### DIFF
--- a/src/Contracts/Auditable.php
+++ b/src/Contracts/Auditable.php
@@ -9,7 +9,7 @@ interface Auditable
     /**
      * Auditable Model audits.
      *
-     * @return \Illuminate\Database\Eloquent\Relations\MorphMany<OwenIt\Auditing\Models\Audit>
+     * @return \Illuminate\Database\Eloquent\Relations\MorphMany<\OwenIt\Auditing\Models\Audit>
      */
     public function audits(): MorphMany;
 
--- a/src/Contracts/Auditable.php
+++ b/src/Contracts/Auditable.php
@@ -3,13 +3,14 @@
 namespace OwenIt\Auditing\Contracts;
 
 use Illuminate\Database\Eloquent\Relations\MorphMany;
+use OwenIt\Auditing\Contracts\Audit as AuditContract;
 
 interface Auditable
 {
     /**
      * Auditable Model audits.
      *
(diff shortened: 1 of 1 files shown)


src/Contracts/Auditable.php
```diff
@@ -9,7 +9,7 @@ interface Auditable
     /**
      * Auditable Model audits.
      *
-     * @return \Illuminate\Database\Eloquent\Relations\MorphMany<OwenIt\Auditing\Models\Audit>
+     * @return \Illuminate\Database\Eloquent\Relations\MorphMany<\OwenIt\Auditing\Models\Audit>
```
erikn69, line 12: I think that must be
```diff
-\OwenIt\Auditing\Models\Audit
+\OwenIt\Auditing\Contracts\Audit
```

aglipanci: That is not correct. The method returns a model and not the contract.

erikn69: Ok, thanks

MortenDHansen:
Haven't seen this as an issue, but we should probably use an import of the contract to be explicit. 

`use OwenIt\Auditing\Contracts\Audit as AuditContract;`

As @erikn69 says, we need to type to the contract. The implementation may vary

aglipanci:
> Haven't seen this as an issue, but we should probably use an import of the contract to be explicit.
> 
> `use OwenIt\Auditing\Contracts\Audit as AuditContract;`
> 
> As @erikn69 says, we need to type to the contract. The implementation may vary

It started to show just today for me as well. I thought of a quick fix, but yeah, looking at it now it makes more sense as this is an interface. 

Let me know if I should update it to that.

aglipanci:
Updated.

anteriovieira:
Thanks @aglipanci

aglipanci:
Any chance you will release a new version anytime soon?

MortenDHansen:
> Any chance you will release a new version anytime soon?

Its tagged now :)

digistorm:
This phpstan error is still persistent in v13.5.1 - The FQN is missing a leading slash.
- Bug report: #843 
- PR: #842

PR #805, opened 2023-03-16
Revert "Update IpAddressResolver.php"

### LINKED ISSUE #804: #743 doesn't respect TrustedProxy settings
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 10.3.3
| Package version   | 13.4.0
| PHP version       | 8.2

#### Actual Behaviour

Insertion fails, as it attempts to insert the ([correctly](https://serverfault.com/questions/846489/can-x-forwarded-for-contain-multiple-ips)) multi-IP forwarding header.

`Illuminate\Database\QueryException: SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type inet: "175.157.86.118, 10.0.16.108" (Connection: pgsql, SQL: insert into "audits" ("old_values", "new_values", "event", "auditable_id", "auditable_type", "user_id", "user_type", "tags", "ip_address", "user_agent", "url", "updated_at", "created_at") values (...  175.157.86.118, 10.0.16.108...)`

#### Expected Behaviour

Laravel Auditing should only use `Request::ip()`. This already respects `X-Forwarded-For` headers internally, **if** the TrustedProxy middleware is correctly configured. That's where this should be handled.

Note that #743 has introduced a security hole, as `X-Forwarded-For` can be modified by a malicious user, resulting in an incorrect IP being logged in the audit trail.

#### Steps to Reproduce

Run Laravel Auditing in a multi-proxy scenario (in our case: nginx --> Docker).

Alternatively, use cURL to spoof the header (`curl -H "X-Forwarded-For: 123.123.123.123" http://example.com`) on an endpoint that will cause an audit log entry. The faked IP should appear in the database. `curl -H "X-Forwarded-For: 123.123.123.123 10.0.0.1" http://example.com` will cause the SQL error.

#### Possible Solutions

Revert change from #743.

erikn69:
Reverts owen-it/laravel-auditing#743
Closes #804
**Info**: https://github.com/owen-it/laravel-auditing/issues/804#issuecomment-1472178550

### DIFF
--- a/src/Resolvers/IpAddressResolver.php
+++ b/src/Resolvers/IpAddressResolver.php
@@ -10,6 +10,6 @@ class IpAddressResolver implements Resolver
 {
     public static function resolve(Auditable $auditable): string
     {
-        return Request::header('x-forwarded-for', Request::ip());
+        return Request::ip();
     }
 }


ceejayoz:
Thanks for creating the PR.

MortenDHansen (approved):

Issue #808, opened 2023-03-23
phpstan internal error in console Method newEloquentBuilder() was not found


expondo:
| Q                 | A
| ----------------- | ---
| Bug?              | no
| New Feature?      | yes
| Framework         | Laravel
| Framework version | 952.4
| Package version   | 13.5.0
| PHP version       | 8.1
| phpstan/phpstan   | 1.10.6

#### Actual Behaviour

When i use Audit::**query** or SomeModel->audits()->**where**() then phpstan throwin exception
`     Error                                                                                                                                                                                        
 -- ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
     Internal error: Internal error: Method newEloquentBuilder() was not found in reflection of class OwenIt\Auditing\Contracts\Audit. in file AuditService.php
     Run PHPStan with -v option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md
     Child process error (exit code 1):
`
this happen after last release with change model to contract: [release diff](https://github.com/owen-it/laravel-auditing/compare/v13.4.0...v13.5.0)
#### Expected Behaviour

Phpstan has no warning when use builder on Audit model

#### Steps to Reproduce

List **all the steps** needed to reproduce the issue you're having.

Make sure to **include code** (affected models, configuration, ...).

Pastebins, screenshots and/or other resources that may help us understand what's going on, will also be appreciated.

In a nutshell, help us, help you.

#### Possible Solutions

AuditContract should be reveiwed.

parallels999:
@MortenDHansen this was alerady fixed on #807

expondo:
Thanks @parallels999 , 
I replied under #807 because I see, that there is some problem still not fixed.

parallels999:
@expondo thanks, Do you know how to improve this part for phpstan?

MortenDHansen:
Well, phpstan was never a part of the workflow, so i have no idea. Im gonna need to add phpstan to our pipeline. Until then i have no idea what was fixed. Its not an issue i would ever see 🤷

expondo:
@parallels999 not yet.
@erikn69 I can check it tomorrow.

fgibaux:
in my code, I use `$model->audits()->...` method and I'm getting the internal error using this package v13.5.0.
when using dev-master, (so with #807), it removes the phpstan internal error but I get normal errors instead:
```
[...]

PR #810, opened 2023-03-24
Test adding PHPStan to mergeflow

MortenDHansen:
see #808


MortenDHansen:
@erikn69 
This is one of those things where i am not sure why we are doing this. 

1. Is it important for our users? - and why? 
2. Is it right to add these changes to the current version? 

I mean, i like types and all that, but it does not seem like our codebase is ready for this.
... or am i just to cautious.

erikn69:
> Is it important for our users? - and why?

It is important to prevent possible bugs

> Is it right to add these changes to the current version?

it passes the tests correctly, It seems to me that there are no breaking changes, but I would feel safer for v14

but I think that it should not be put in the workflow, I am going to do a PR to show you

MortenDHansen:
> but I think that it should not be put in the tests workflow, I am going to do a PR to show you

I like that version better :)

MortenDHansen:
@erikn69 - changed to v14 branch. If you approve, well make it so :)

erikn69:
Hi @MortenDHansen I noticed that the tests are not running, something broke or it is because it is the dev branch

PR #811, opened 2023-03-24
Add larastan analyse

### LINKED ISSUE #808: phpstan internal error in console Method newEloquentBuilder() was not found

erikn69:
Complement of #810
Closes #808
seems like this make phpstan happy

### DIFF
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -15,8 +15,13 @@ class Database implements AuditDriver
     public function audit(Auditable $model): ?Audit
     {
         $implementation = Config::get('audit.implementation', \OwenIt\Auditing\Models\Audit::class);
+        $callback = [$implementation, 'create'];
 
-        return call_user_func([$implementation, 'create'], $model->toAudit());
+        if(! is_callable($callback)) {
+            throw new \UnexpectedValueException("Method config('audit.implementation')::create() does not exist.");
+        }
+
+        return call_user_func($callback, $model->toAudit());
     }
 
     /**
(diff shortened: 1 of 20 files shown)


src/Drivers/Database.php
```diff
@@ -26,14 +31,14 @@ public function prune(Auditable $model): bool
     {
         if (($threshold = $model->getAuditThreshold()) > 0) {
             $forRemoval = $model->audits()
+                ->select('id')
                 ->latest()
-                ->get()
-                ->slice($threshold)
-                ->pluck('id');
+                ->offset($threshold)->limit((int) '18446744073709551615')
```
MortenDHansen, line 36: What is this? :)

erikn69: Ok, larastan says that we are bringing unnecessary data to memory.
Example, if we have `$threshold = 1000` and audits are 1001, we brings to memory 1001 colection and slide just last one
maybe it would be better to open another PR to discuss this change, i can revert this and set an exception for phpstan
I imagine this goes for V14

erikn69: Other better way just delete, without checking empty
```php
return $model->audits()
      ->select('id')
      ->latest()
      ->offset($threshold)->limit((int) '18446744073709551615')
      ->delete() > 0;
```

erikn69: This change was added here [_(12ded01)_](https://github.com/owen-it/laravel-auditing/commit/12ded011b140140fea3f8ae49ccc45493aff9eff ), any reason to do it like this?

PR #814, opened 2023-03-28
[V14] Use anonymous migrations (for L8+)

erikn69:
Since Laravel 8.x it is possible to use anonymous migrations
It's time to change migrations


MortenDHansen (approved):

MortenDHansen:
I agree!! - its about time :)

PR #816, opened 2023-03-29
Remove publishing on lumen

erikn69:
Complement of https://github.com/owen-it/laravel-auditing.com/pull/37

### DIFF
--- a/src/AuditingServiceProvider.php
+++ b/src/AuditingServiceProvider.php
@@ -51,13 +51,17 @@ class AuditingServiceProvider extends ServiceProvider implements DeferrableProvi
      */
     private function registerPublishing()
     {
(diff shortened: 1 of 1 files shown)


erikn69:
@MortenDHansen hi, 
what do you think of [irazasyed/larasupport](https://github.com/irazasyed/larasupport)?
is it a good replacement for [laravelista/lumen-vendor-publish](https://github.com/laravelista/lumen-vendor-publish)?
If this is the case, we should reverse this PR and modify only the documentation

MortenDHansen:
Honestly, i don't know. I mean, I've never even installed lumen so i wouldn't know either way.

erikn69:
I did test it, it works

Issue #828, opened 2023-05-16
Audit Formatting Doesn't Include Attribute Accessors


hamedghaderi:
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no|yes
| Framework         | Laravel|Lumen
| Framework version | 10.8.0
| Package version   | 13.5.0
| PHP version       | 8.2

#### Actual Behaviour

Currently, when the Audit functionality formats data, it fails to detect and include attribute accessors of type "Attribute" during the formatting process. As a result, attribute accessors are not considered when formatting the data.

#### Expected Behaviour

It is expected that the Audit functionality should include attribute accessors of type "Attribute" during the data formatting process. This would ensure that any custom formatting defined through attribute accessors is properly considered.

#### Steps to Reproduce
1. Assume we have a model `Foo` with a JSON attribute, `bar`, casted as an `ArrayObject`:

```php
class Foo extends Model implements Auditable 
{
  protected $casts = [
     'bar' => AsArrayObject::class,
  ];
}

```

2. Before fetching the bar attribute, we add custom formatting using an attribute accessor:
```php
public function bar(): Attribute
{
  return Attribute::make(
    get: fn ($value) => // Some custom formatting 
  )
}
```

3. The Audit functionality currently ignores this custom formatting because it only checks for attribute accessors starting with get, such as getBarAttribute.

<img width="724" alt="Screenshot 2023-05-16 at 11 33 46 AM" src="https://github.com/owen-it/laravel-auditing/assets/45762462/0f86bad1-1677-4308-8439-3dc84d262b8a">

PR #829, opened 2023-05-16
Include Attribute Mutator on Audit Formatting

hamedghaderi:
## Issue

This pull request addresses the issue of the Audit functionality in the Laravel framework failing to include attribute accessors  during the formatting process. 

## Solution

To solve this issue, I have made modifications to the Audit functionality. The changes now include attribute accessors  when formatting the data. This ensures that any custom formatting defined through attribute accessors is properly considered.

Related Issue #828 

Checklist
- [x] Tests have been added to cover the changes.
- [ ] Documentation has been updated to reflect the changes.

### DIFF
--- a/tests/Models/Article.php
+++ b/tests/Models/Article.php
@@ -12,6 +12,8 @@ class Article extends Model implements Auditable
     use \OwenIt\Auditing\Auditable;
     use SoftDeletes;
 
+    protected $laravel_version;
+
     /**
      * {@inheritdoc}
      */
@@ -72,8 +74,17 @@ class Article extends Model implements Auditable
     public function content(): Attribute
     {
         return new Attribute(
-            get: fn ($value) => $value,
-            set: fn ($value) => ucwords($value),
+            function ($value) { return $value; },
+            function ($value) { return ucwords($value); }
         );
+    }
+
+    public static function contentMutate($value)
+    {
+        if (! method_exists(self::class, 'hasAttributeMutator')) {
+            return $value;
+        }
+
+        return ucwords($value);
     }
 }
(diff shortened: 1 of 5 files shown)


parallels999:
Closes #828

anteriovieira:
After we have tests and doc I can merge it. Would someone be able to help us?

hamedghaderi:
@anteriovieira The change aligns with standard Laravel behavior, so there's no need for documentation updates. While it may seem like a new feature, it's essentially a bug-fix for Laravel's attribute mutators.

MortenDHansen:
Thanks @hamedghaderi. We do however run into an issue. The named parameters in the new laravel style would only work from PHP 8.0 and v13 support 7.4

On v8.0 and upo, the test fails, but i haven't looked into why yet

tests/Models/Article.php
```diff
@@ -62,4 +63,17 @@ public function getTitleAttribute(string $value): string
...
+
+    /**
+     * Uppercase Content accessor.
+     *
+     * @return Attribute
+     */
+    public function content(): Attribute
+    {
+        return new Attribute(
+            get: fn ($value) => $value,
```
MortenDHansen, line 75: This will break with the testsuite. The package in v13 needs to be compatible with php7.4

erikn69: https://github.com/hamedghaderi/laravel-auditing/pull/1

erikn69: Fixed with `method_exists($model, 'hasAttributeMutator')`

hamedghaderi: Thanks to @erikn69, the new patch should fix the issue.

MortenDHansen (approved):

PR #838, opened 2023-06-08
Update return value of Auditor->execute so that it works for PHP Version 8.1.19

weirdlooop:
This error is returned if return type void is not provided using  PHP Version > 8.1.19:

Declaration of OwenIt\Auditing\Auditor::execute(OwenIt\Auditing\Contracts\Auditable $model) must be compatible with OwenIt\Auditing\Contracts\Auditor::execute(OwenIt\Auditing\Contracts\Auditable $model): void

### DIFF
--- a/src/Auditor.php
+++ b/src/Auditor.php
@@ -55,7 +55,7 @@ class Auditor extends Manager implements Contracts\Auditor
     /**
      * {@inheritdoc}
      */
-    public function execute(Auditable $model)
+    public function execute(Auditable $model): void
     {
         if (!$model->readyForAuditing()) {
             return;


jgrandar:
Similar problem with PHP 8.2.6, got:
local.ERROR: Declaration of OwenIt\Auditing\Auditable::auditDetach(string $relationName, $ids = null, $touch = true) must be compatible with App\Models\BaseModel::auditDetach($relationName, $ids) {"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Declaration of OwenIt\\Auditing\\Auditable::auditDetach(string $relationName, $ids = null, $touch = true) must be compatible with App\\Models\\BaseModel::auditDetach($relationName, $ids) at /usr/share/nginx/html/vendor/owen-it/laravel-auditing/src/Auditable.php:674)
[stacktrace]

PR #839, opened 2023-06-17
fix FQCN and add eloquent method to interface

mintbridge:
it stops phpstan from reporting this:
```
Internal error: Internal error: Method newEloquentBuilder() was not found in reflection of class
     OwenIt\Auditing\Contracts\Audit. in file
     Controller.php
     Run PHPStan with -v option and post the stack trace to:
     https://github.com/phpstan/phpstan/issues/new?template=Bug_report.yaml
     Child process error (exit code 1):
```

### DIFF
--- a/src/Contracts/Audit.php
+++ b/src/Contracts/Audit.php
@@ -5,6 +5,14 @@ namespace OwenIt\Auditing\Contracts;
 interface Audit
 {
     /**
+     * Create a new Eloquent query builder for the model.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @return \Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|static
+     */
+    public function newEloquentBuilder($query);
+    
+    /**
      * Get the current connection name for the model.
      *
      * @return string|null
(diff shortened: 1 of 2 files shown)


src/Contracts/Audit.php
```diff
@@ -4,6 +4,14 @@
 
 interface Audit
 {
+    /**
+     * Create a new Eloquent query builder for the model.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @return \Illuminate\Database\Eloquent\Builder<\Illuminate\Database\Eloquent\Model>|static
+     */
+    public function newEloquentBuilder($query);
```
parallels999, line 13: You can't change an interface like this, that could become a breaking change

If it's only for PhpStan try adding this at the start of the interface

```php
/**
 * @mixin \Illuminate\Database\Eloquent\Model
 */
interface Audit
{
```

parallels999: Also you could add 
```php
/*
 * @method static \Illuminate\Database\Query\Builder newEloquentBuilder(\Illuminate\Database\Query\Builder  $query)
 */
interface Audit
{
```

Issue #562, opened 2019-11-12
Run model auditing observer as job


prodigy7:
| Q                 | A
| ----------------- | ---
| Bug?              | no
| New Feature?      | yes
| Framework         | Laravel
| Framework version | 6.x
| Package version   | 9.x
| PHP version       | 7.x

#### Actual Behaviour

Currently, when a model is changed, the observer was directly triggered.

#### Expected Behaviour

I wish a optional configurable behaviour, that a audit record is written by a job. Problem in my case is, that I am working with a big amount of data. I want audit any change but it's ok, when it's done some seconds later. In this situation I want win some time by move the auditing process to background.

#### Steps to Reproduce

Take existing model, change something and save.

#### Possible Solutions

Implement current observer auditing code into own class and call it directly or with job

applyACS:
I'm having the same issue. I'd like to run audit as a job.

MortenDHansen:
I see the value of the idea. A bunch of operations could be pushed back doing this. Maybe let the observer trigger a new event that could queue for doing the actual work. 

We will look into it

parallels999:
https://github.com/owen-it/laravel-auditing/blob/22682dd5c800aab90ef261530098625e26a66709/config/audit.php#L181-L186

PR #845, opened 2023-08-22
Queued Auditing

Orrison:
Per discussions such as #562, this sets up auditing to be dispatched on the Queue.

By default, the listener will go onto the `sync` connection. So if no configuration is done, the package will work as it did before, auditing immediately in the current request.

Once you configure the `audit.queue.connect` to be something other than `sync`, the audits will then be dispatched to be processed asynchronously.

Customization of the queue it is put on, as well as the delay, is also possible.

It is set up similarly to the `Auditing/Audited` flow. An Event (`DispatchingAudit`) is fired where you can prevent the auditing dispatch from happening by having a listener return `false`.
And there is the ability to swap out the `ProcessDispatchAudit` Listener if the end user needs to.


Orrison:
Closed in favor of different fork location

PR #851, opened 2023-08-24
Make UrlResolver resolve the proper command line used instead of "console"

tigitz:
Hello,

I've had ambiguous audits in the past where we couldn't determinate from which of our scheduled command these modifications were coming from because everything is "console", so the debugging experience were not ideal.

To improve the situation I implemented a custom way to retrieve the proper command line used while generating the audit, this works great for us. 

image

So I decided to contribute it.

*Note*: Since the url metadata is now non deterministic in tests because of various possible ways to run the tests (arguments, options, paths, etc.), I've decided to expose the command line resolving logic to make use of it in tests that are not meant to test this feature specifically. 
And then introduce a test that is meant to run with a specific command line and thus expect a deterministic output. This one is actually testing the logic.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -58,4 +58,6 @@ jobs:
           composer update --prefer-dist --no-interaction --no-suggest
 
       - name: Execute tests
-        run: vendor/bin/phpunit
+        run: |
+          vendor/bin/phpunit
+          vendor/bin/phpunit tests/Unit/AuditTest.php --group command-line-url-resolver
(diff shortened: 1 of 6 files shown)


MortenDHansen (approved):

MortenDHansen:
I like that idea! 

What say you @erikn69 / @anteriovieira ?

erikn69:
I think it's a great idea, we would have more detailed information where the audit comes from

tigitz:
Great to see everyone on board with the feature :slightly_smiling_face: . Should I expect it to be merged for the next release ? Thanks

PR #866, opened 2023-10-04
Refactor auditSync, auditDetach, auditAttach methods

### LINKED ISSUE #859: auditSync - writes an empty array when there is no data change
| Q                 | A
| ----------------- | ---
| Bug?              | yes
| New Feature?      | no
| Framework         | Laravel
| Framework version | 10.10
| Package version   | 13
| PHP version       | 8.1

#### Actual Behaviour

Empty array - [] when using auditSync.

When there is no change in auditSync, an empty array is written to the database and fills the table with unnecessary records.
Database:
event: sync
old_values: []
new_values: []

Relationship: $this->belongsToMany(User::class)->withTimestamps();

$image->auditSync('users', $validated['users']);

#### Expected Behaviour

When there is no relationship change and auditSync is executed, do not write anything to the database - it is currently writing an empty array.

#### Steps to Reproduce

Model Relationship: $this->belongsToMany(User::class)->withTimestamps();
View: $validated['users'] - array of ids - should not be changed, just call update again in the controller 
Controller:$image->auditSync('users', $validated['users']);

#### Possible Solutions

Add a check that the old array of ids and new array of ids data match.

erikn69:
This PR keeps the current functionality, just better readability, PhpDoc fixes

@MortenDHansen Hi,
I want to make a change in the auditing of these relationship methods, I added `Collection::diff` o my local copy, so that only the  `added/removed` records in the pivot are audited, and in the case that there are no changes and `'audit.empty_values'` is false the audit would not be recorded
That change depends on this being accepted.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -647,6 +647,7 @@ trait Auditable
      * @param mixed $id
      * @param array $attributes
      * @param bool $touch
+     * @param array $columns
      * @return void
      * @throws AuditingException
      */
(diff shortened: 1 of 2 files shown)


gisostallenberg:
@erikn69 Could you look into https://github.com/owen-it/laravel-auditing/issues/869 to see what can be improved there?

MortenDHansen (approved):

MortenDHansen:
Looks about right. :)

parallels999 (approved):

erikn69:
> Looks about right. :)

Ok, I will wait for it to be merged to add another PR and discuss the new functionality, I don't like refactoring something and adding features at the same time

PR #870, opened 2023-10-06
Lets not type stuff that is not supposed to be typed

MortenDHansen:


erikn69:
Sorry, I hadn't seen it and I opened another one #872
> why don't you use `Squash & Merge`? The `commit history` is easier to follow, it is more linear.

MortenDHansen:
> Sorry, I hadn't seen it and I opened another one #872
> 
> > why don't you use `Squash & Merge`? The `commit history` is easier to follow, it is more linear.

Geez! - right! No good reason for that other than me not thinking about it

PR #871, opened 2023-10-10
Read Audit implementation from MorphMany

gisostallenberg:
To allow a different Audit implementation **per model** the implementation should be deducted from the MorphMany given by the Auditable::audits method.

This will allow having a main Audit implementation (which defaults to `OwenIt\Auditing\Models\Audit`), but also have a different implementation once the Auditable::audits method is overwritten.

Example:
```php
class MyAuditable extends Model implements AuditableContract
{
    use Auditable;

    /**
     * @return MorphMany<AuditOfMyAuditableModel>
     */
    public function audits(): MorphMany
    {
        return $this->morphMany(
            AuditOfMyAuditableModel::class,
            'auditable',
        );
    }
```

### DIFF
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -14,9 +14,7 @@ class Database implements AuditDriver
      */
     public function audit(Auditable $model): ?Audit
     {
-        $implementation = Config::get('audit.implementation', \OwenIt\Auditing\Models\Audit::class);
-
-        return call_user_func([$implementation, 'create'], $model->toAudit());
+        return call_user_func([$model->audits()->getModel()::class, 'create'], $model->toAudit());
     }
 
     /**


MortenDHansen:
It is a good idea! - I do need some test coverage of the feature though.

parallels999:
@gisostallenberg hi, rebase it

src/Drivers/Database.php
```diff
@@ -14,9 +14,7 @@ class Database implements AuditDriver
      */
     public function audit(Auditable $model): ?Audit
     {
-        $implementation = Config::get('audit.implementation', \OwenIt\Auditing\Models\Audit::class);
-
-        return call_user_func([$implementation, 'create'], $model->toAudit());
+        return call_user_func([$model->audits()->getModel()::class, 'create'], $model->toAudit());
```
parallels999, line 17: Use `get_class`
```php
return call_user_func([get_class($model->audits()->getModel()), 'create'], $model->toAudit());
```

erikn69:
Fix on https://github.com/owen-it/laravel-auditing/commit/a08c673e31bc4d55b6f5b5dd8cc9ab9085022d15

PR #875, opened 2023-10-18
Use `https` for the homepage on the `composer.json` file

majidalaeinia:
Use 'https' instead of  'http' for the homepage


angeljqv:
`http` redirects to `https`, works the same to me

MortenDHansen (approved):

PR #881, opened 2023-12-15
Make queued audits feature opt-in

erikn69:


MortenDHansen (approved):

MortenDHansen:
Seems like a good fix. Im not sure if there is something we should take into account before releasing it? (@Orrison ? )

Orrison:
This should do the trick I believe.

Orrison (approved):

parallels999:
I didn't know how to solve my problem, but with this at least I don't have to go back to a previous version

parallels999 (approved):

angeljqv (approved):

cosmastech:
Isn't this technically making it "opt-out" since the default is now to queue where it wasn't previously?

PR #882, opened 2023-12-18
Make queued auditing opt-in

cosmastech:
The previous PR #881 made the default behavior opt-out, while I think we actually want it to be opt-in to avoid breaking changes.

For some more context, I am updating composer packages and our test cases were failing that were concerned with auditing. After some digging, I realized I had to copy over the updates to config/audit.php and ensure that we had `'audit.queue.enabled'` set to `false`.

With the change proposed in this PR, there's no need to copy over the configuration (which isn't always obvious), it will just work exactly as it did previously without updating the configuration.

### DIFF
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -105,7 +105,7 @@ class AuditableObserver
             return;
         }
 
-        if (!Config::get('audit.queue.enable', true)) {
+        if (!Config::get('audit.queue.enable', false)) {
             return Auditor::execute($model);
         }
 


MortenDHansen (approved):

MortenDHansen:
Agreed 👍

spaantje:
Please note that this change does not prevent the job from being queued. Because of the change in the main config it still queues the job when the config is published and not updated. You should also update the root config to `false` which was changed/added in https://github.com/owen-it/laravel-auditing/pull/881 

(config/audit.php > queue.enable)

cangelis:
> Please note that this change does not prevent the job from being queued. Because of the change in the main config it still queues the job when the config is published and not updated. You should also update the root config to `false` which was changed/added in #881
> 
> (config/audit.php > queue.enable)

100%. The default value should also be `false` in `config/audit.php`

There is a breaking change currently re queues and this PR doesn't really prevent the issue happening.

See: https://github.com/owen-it/laravel-auditing/issues/890

PR #899, opened 2024-02-06
feat (Auditable): customized relationship for Attach, Detach and Sync

wilianx7:
Hello guys. 

In my API I have a situation where I need to provide a BelongsToMany relationship with some conditions to Audit Attach, Detach and Sync.

This situation occurs because my pivot table has many pivot columns that need to be conditioned when I do an Attach/Detach/Sync.

Nowadays I can't do this, as these methods only accept the name of the relationship (without conditions). So I made this PR with a new optional parameter `$relationship`.

Example use case:

image

In this case, I just want to sync the serviceTypes relationship in my pivot table, where the pivot columns match the respective company_id and tenancy_id values. The other data in the pivot table remained unchanged.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -664,54 +664,54 @@ trait Auditable
     */
 
     /**
-     * @param string|\Closure<BelongsToMany> $relation
+     * @param string $relationName
      * @param mixed $id
      * @param array $attributes
      * @param bool $touch
      * @param array $columns
+     * @param \Closure|null $callback
      * @return void
      * @throws AuditingException
      */
-    public function auditAttach($relation, $id, array $attributes = [], $touch = true, $columns = ['*'])
+    public function auditAttach(string $relationName, $id, array $attributes = [], $touch = true, $columns = ['*'], $callback = null)
     {
-        if (is_string($relation)) {
-            $this->validateRelationshipMethodExistence($relation, 'attach');
-        } else {
-            $this->validateBelongsToManyRelationshipClosure($relation);
+        $this->validateRelationshipMethodExistence($relationName, 'attach');
+
+        $relationCall = $this->{$relationName}();
+
+        if ($callback instanceof \Closure) {
+            $this->applyClosureToRelationship($relationCall, $callback);
         }
 
-        $relationCall = is_string($relation) ? $this->{$relation}() : $relation();
         $old = $relationCall->get($columns);
         $relationCall->attach($id, $attributes, $touch);
         $new = $relationCall->get($columns);
-
-        $relationName = is_string($relation) ? $relation : $relationCall->getRelationName();
 
         $this->dispatchRelationAuditEvent($relationName, 'attach', $old, $new);
     }
 
     /**
-     * @param string|\Closure<BelongsToMany> $relation
+     * @param string $relationName
      * @param mixed $ids
      * @param bool $touch
      * @param array $columns
+     * @param \Closure|null $callback
      * @return int
      * @throws AuditingException
      */
-    public function auditDetach($relation, $ids = null, $touch = true, $columns = ['*'])
+    public function auditDetach(string $relationName, $ids = null, $touch = true, $columns = ['*'], $callback = null)
     {
-        if (is_string($relation)) {
-            $this->validateRelationshipMethodExistence($relation, 'detach');
-        } else {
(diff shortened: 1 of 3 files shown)


src/Auditable.php
```diff
@@ -688,18 +691,20 @@ public function auditAttach(string $relationName, $id, array $attributes = [], $
...
      * @param bool $touch
      * @param array $columns
+     * @param BelongsToMany|null $relationship
      * @return int
      * @throws AuditingException
      */
-    public function auditDetach(string $relationName, $ids = null, $touch = true, $columns = ['*'])
+    public function auditDetach(string $relationName, $ids = null, $touch = true, $columns = ['*'], $relationship = null)
     {
         if (!method_exists($this, $relationName) || !method_exists($this->{$relationName}(), 'detach')) {
```
parallels999, line 700: I think it would be better to add a closure. this change make `$relationName` useless

parallels999:
Also needs unit tests

src/Auditable.php
wilianx7: I think the relationship name is necessary because it is used in the dispatchRelationAuditEvent method. Do you have any suggestions on how to get the relationship name another way while maintaining the functionality I'm trying to add?

However, the flaws you pointed out really exist.

parallels999: >Do you have any suggestions on how to get the relationship name another way while maintaining the functionality I'm trying to add?

>I think it would be better to add a closure

Already answered, a closure, Laravel uses closures for these cases, that will maintain consistency

parallels999:
That's not what I meant by Laravel style. 
I mean something like this
```php
/* @param array $attributes
 * @param bool $touch
 * @param array $columns
 * @param \Closure|null $callback 
 * @return void
 * @throws AuditingException
 */
public function auditAttach(string $relationName, $id, array $attributes = [], $touch = true, $columns = ['*'], $callback = null)
{
    if (!method_exists($this, $relationName) || !method_exists($this->{$relationName}(), 'attach')) {
        throw new AuditingException('Relationship ' . $relationName . ' was not found or does not support method attach');
    }

    $relationCall = $this->{$relationName}();

    if ($callback instanceof \Closure) {
        $relationCall = $callback($relationCall);
    }

    $old = $relationCall->get($columns);
    $relationCall->attach($id, $attributes, $touch);
    $new = $relationCall->get($columns);
    $this->dispatchRelationAuditEvent($relationName, 'attach', $old, $new);
}
```

wilianx7:
I understand. However, in the way you are suggesting, the method call becomes much more verbose taking into account the addition of an optional parameter at the end, which forces the passage of all other parameters so that Closure can be used.

The implementation I made is more dynamic, as it does not require an additional parameter and allows the method to be used from a Closure or the name of the relationship (as it is today).

Suggested implementation:

`$model->auditAttach('my_existing_relation', ['ids'], null, null, null, fn ($rel) => $rel->wherePivot());`

My implementation:

`$model->auditAttach(fn () => $rel->wherePivot(), ['ids']);`

`$model->auditAttach('my_existing_relation', ['ids']);`

What would be the negative point of my implementation?

parallels999:
>What would be the negative point of my implementation?

It seems that you did not read my previous comment on your tests, I will repeat it again:
```php
$article1 = factory(Article::class)->create();
$article2 = factory(Article::class)->create();

$closure = function () use ($article1) { return $article1->categories(); };
$article2->auditAttach($closure, $firstCategory);
```
Here, where `$firstCategory` is attached? `$article1` or `$article2`?

parallels999:
> I made is more dynamic, as it does not require an additional parameter

[blog/php-8-named-arguments#why-named-arguments?](https://stitcher.io/blog/php-8-named-arguments#why-named-arguments?)
You could use named arguments
```php
$model->auditAttach(
    relationName: 'my_existing_relation', 
    callback: fn ($rel) => $rel->wherePivot()
);
```

wilianx7:
> Here, where `$firstCategory` is attached? `$article1` or `$article2`?

Sorry, I didn't see any comments in my tests. But you're right, in this situation my implementation will fail.

I'll do it the way you suggested ASAP.

wilianx7:
@parallels999 could you review, please?

parallels999:
LGTM 👍
But @MortenDHansen review is needed

wilianx7:
@MortenDHansen could you review, please?

PR #904, opened 2024-02-28
Support 404labfr/laravel-impersonate

jose123v:
Closes #903

>Using [404labfr/laravel-impersonate](https://github.com/404labfr/laravel-impersonate) in [owen-it/laravel-auditing](https://github.com/owen-it/laravel-auditing) saves the impersonated user instead of the impersonating user

### DIFF
--- a/src/Resolvers/UserResolver.php
+++ b/src/Resolvers/UserResolver.php
@@ -13,6 +13,15 @@ class UserResolver implements \OwenIt\Auditing\Contracts\UserResolver
      */
     public static function resolve()
     {
+        // supports https://github.com/404labfr/laravel-impersonate
+        if (app()->bound('impersonate')) {
+            /** @var \Lab404\Impersonate\Services\ImpersonateManager */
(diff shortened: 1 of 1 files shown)


willpower232 (changes requested):
Technically speaking, one has the ability to provide a custom user resolver in the configuration so this would be better as its own package which relies on owen-it/laravel-auditing and perhaps extends the base user resolver so it can still be used via similar code.

Also if this package starts supporting this impersonation library, it would probably end up supporting other libraries that change the user and that would require extra time and testing from the maintainers.

cesarreyes3:
It may take some time until you realize this problem, it would be better to have it

erikn69:
I have my doubts still, 
it could help to get the real executor of the action in the case of that package, and if that package does not exist it should not affect the performance and it seems fine to me

calebdw:
I use both packages and my audits table has both `user` and `impersonator` columns

Issue #596, opened 2020-07-20
Model changes not being traced in case of Array of Objects in case of MongoDB


Y003:
Hi,

I'm using the version no. 9.0 for auditing with laravel version 5.1 & mongoDB as the database, in case of an array of object in one of my model, changes in this particular array field are not being captured. Below is the example array:

"spoc" : [ 
        {
            "first_name" : "xxx",
            "middle_name" : "xx",
            "last_name" : "xx",
            "contact" : "xxxxxxxxxx",
            "email" : "x.x@x.com",
            "default" : true
        }, 
        {
            "first_name" : "test",
            "middle_name" : "test",
            "last_name" : "test",
            "contact" : "1234567890",
            "email" : "test.test@test.com"
        }
    ]

Please help me get through this problem.
Thanks.

aizaztag:
Having same issue have you fixed that one ,  I am unable to see array in old/new values

aizaztag:
After searching and struggling what i got is 
Non-scalar values, such as arrays and objects, can be saved using the JSON data type. However, if the data is too large, it may exceed the maximum size of the JSON column in your database.

If you are unable to save non-scalar values using the package, you may consider creating a custom audit driver that uses a different data type to store the old and new values of the models. This will allow you to store non-scalar data in a way that is compatible with your database.

Another solution is to use a package that support saving non-scalar data by converting them to string representation before saving it.

PR #915, opened 2024-03-19
Add config to allow records array values

rizkypujiraharja:
In projects utilizing MongoDB as the database, data is commonly stored in the form of arrays and objects. However, non-scalar values are not allowed to be recorded in this packages. 

Therefore, I have added a new config to provide users with the option to audit array values.

Issue: #596

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -108,6 +108,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Allowed Array Values
+    |--------------------------------------------------------------------------
+    |
+    | Should the array values be audited?
+    |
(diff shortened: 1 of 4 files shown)


MortenDHansen:
Please add a test which shows the difference ensuring flipping the config works, and not implementing the change does not break anything :)

oorizky:
> Please add a test which shows the difference ensuring flipping the config works, and not implementing the change does not break anything :)

 I have already created the test as requested.

rizkypujiraharja:
> Please add a test which shows the difference ensuring flipping the config works, and not implementing the change does not break anything :)

I've added a test, please review again :)

MortenDHansen (approved):

PR #917, opened 2024-03-21
Add `withoutAuditing()` callback method to temporarily disable auditing for one class

derekmd:
## Summary

Similar to Laravel's `Model::withoutEvents()` method, allow running Eloquent queries on a given model class inside a callback:

```php
User::withoutAuditing(function () use ($user, $inputs, $additional) {
    // no audit
    $user->update($inputs);

    // this _will_ be audited if the Profile class is Auditable
    $user->profile()->updateOrCreate([], $additional);
});
```

Only the `{class}::withoutAuditing()` class will have auditing disabled, so additional model classes need their own call. e.g.,

```php
User::withoutAuditing(fn () => $user->update($inputs));
Profile::withoutAuditing(fn () => $user->profile()->updateOrCreate([], $additional));
```

---

1. It's wrapped in `try` / `finally` to handle restoring auditing after Eloquent possibly throws `QueryException`.
2. If auditing is already disabled, that setting will be kept. i.e., don't call `enableAuditing()`

## Links

* https://laravel.com/docs/11.x/eloquent#muting-events
* https://github.com/laravel/framework/blob/6c9cbc9ba0437bcfe54609ae0af312feefc7e309/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php#L426-L441
* https://github.com/laravel/framework/blob/e307d111c22066ef614a112664724b4db4956968/src/Illuminate/Support/Traits/Localizable.php#L16 Another example using the same design pattern.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -537,6 +537,26 @@ trait Auditable
     }
 
     /**
+     * Execute a callback while auditing is disabled.
+     *
+     * @param callable $callback
+     *
+     * @return mixed
+     */
+    public static function withoutAuditing(callable $callback)
+    {
+        $auditingDisabled = static::$auditingDisabled;
+
+        static::disableAuditing();
+
+        try {
+            return $callback();
+        } finally {
+            static::$auditingDisabled = $auditingDisabled;
+        }
+    }
+
+    /**
      * Determine whether auditing is enabled.
      *
      * @return bool
--- a/tests/Functional/AuditingTest.php
+++ b/tests/Functional/AuditingTest.php
@@ -433,6 +433,29 @@ class AuditingTest extends AuditingTestCase
 
     /**
      * @test
+     */
+    public function itDisablesAndEnablesAuditingBackAgainViaWithoutAuditingMethod()
+    {
+        // Auditing is enabled by default
+        $this->assertFalse(Article::$auditingDisabled);
+
+        Article::withoutAuditing(function () {
+            factory(Article::class)->create();
+        });
+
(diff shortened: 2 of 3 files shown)


willpower232 (commented):
I don't hate this but I must ask, since this package makes use of Laravel events, could you not make use of `::withoutEvents()` or `->saveQuietly()` to achieve the same thing?

derekmd:
You can but that will also prevent other event listeners from being invoked. This method would allow only disabling audit callbacks.

anteriovieira (approved):

anteriovieira:
Could someone please update the documentation? You could follow the idea from the Laravel documentation.

https://laravel.com/docs/11.x/eloquent#muting-events

willpower232:
@derekmd would you like the honours? https://github.com/owen-it/laravel-auditing.com/ perhaps something under the advanced section?

derekmd:
https://github.com/owen-it/laravel-auditing.com/pull/44

without-events-docs

erikn69 (approved):

erikn69:
@derekmd hi, this give me test errors
```php
$result = Article::withoutAuditing(function () {
    $this->assertTrue(Article::$auditingDisabled);
    $this->assertFalse(ArticleExcludes::$auditingDisabled); //here, is true instead of false, why?
```

erikn69:
Thanks,  I already noticed it, #937

erikn69:
maybe another idea,
globally disable audits on all models within the callback block, maybe it could be done with a second argument

PR #921, opened 2024-04-11
feat (Auditable): customized relationship for Attach, Detach and Sync

wilianx7:
Hello guys. 

In my API I have a situation where I need to provide a BelongsToMany relationship with some conditions to Audit Attach, Detach and Sync.

This situation occurs because my pivot table has many pivot columns that need to be conditioned when I do an Attach/Detach/Sync.

Nowadays I can't do this, as these methods only accept the name of the relationship (without conditions). So I made this PR with a new optional parameter `$relationship`.

Example use case:

image

In this case, I just want to sync the serviceTypes relationship in my pivot table, where the pivot columns match the respective company_id and tenancy_id values. The other data in the pivot table remained unchanged.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -664,54 +664,54 @@ trait Auditable
     */
 
     /**
-     * @param string|\Closure<BelongsToMany> $relation
+     * @param string $relationName
      * @param mixed $id
      * @param array $attributes
      * @param bool $touch
      * @param array $columns
+     * @param \Closure|null $callback
      * @return void
      * @throws AuditingException
      */
-    public function auditAttach($relation, $id, array $attributes = [], $touch = true, $columns = ['*'])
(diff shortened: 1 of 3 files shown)


parallels999:
- Any difference with #899?

wilianx7:
It's because I have published the master branch of my fork on Packgist. So I created another branch for this PR.

src/Auditable.php
```diff
@@ -365,11 +366,11 @@ public function transformAudit(array $data): array
      *
      */
     protected function resolveUser()
-    {   
+    {
         if (! empty($this->preloadedResolverData['user'] ?? null)) {
             return $this->preloadedResolverData['user'];
         }
-
+
```
willpower232, line 373:
```suggestion

```

PR #930, opened 2024-05-22
Possible bug in auditSync

tricki:


tests/Functional/AuditingTest.php
```diff
@@ -620,13 +620,16 @@ public function itWillAuditSync()
...
+        $audit = $article->audits()->where('event', 'sync')->latest()->first();
         $no_of_audits_after = Audit::where('auditable_type', Article::class)->count();
         $categoryAfter = $article->categories()->first()->getKey();
 
         $this->assertSame($firstCategory->getKey(), $categoryBefore);
         $this->assertSame($secondCategory->getKey(), $categoryAfter);
         $this->assertNotSame($categoryBefore, $categoryAfter);
         $this->assertGreaterThan($no_of_audits_before, $no_of_audits_after);
+        $this->assertCount(1, $audit->old_values['categories']);
+        $this->assertCount(2, $audit->new_values['categories']);
```
erikn69, line 632: `sync` is expected to delete the entire relationship and re-attach it

https://laravel.com/docs/11.x/eloquent-relationships#syncing-associations

So after sync, `old_values` has 1, and `new_values` has 1

tricki: Yeah, I just saw that. I've pushed a fix and a new test which illustrates my problem better. I wasn't expecting you to be this fast 😅
Here's the corresponding issue with more explanation: https://github.com/owen-it/laravel-auditing/issues/931

tests/Functional/AuditingTest.php
```diff
@@ -620,13 +620,46 @@ public function itWillAuditSync()
...
+        $firstCategory = factory(Category::class)->create();
+        $secondCategory = factory(Category::class)->create();
+        $thirdCategory = factory(Category::class)->create();
+        $article = factory(Article::class)->create();
+
+        $article->categories()->attach($firstCategory);
+
+        $no_of_audits_before = Audit::where('auditable_type', Article::class)->count();
+
+        $article->auditSync('categories', [$secondCategory->getKey(), $thirdCategory->getKey()]);
```
willpower232, line 650: are you expecting this to **add** two categories or **set** the whole relation to just these two categories?

tests/Functional/AuditingTest.php
erikn69: the change I made is how I understood your problem

tests/Functional/AuditingTest.php
tricki: To **set** it to the two categories. I'm also expecting that "old" contains the first category (i.e. the same as `$article->categories` before the sync), and "new" the second and third (i.e. the same as `$article->categories` after the sync).

willpower232: Makes sense thanks, looks like @erikn69 tweaked the test so it fails now so hopefully is on the case 🙏

parallels999:
It seems it was a misunderstanding: https://github.com/owen-it/laravel-auditing/issues/931#issuecomment-2126371800

PR #932, opened 2024-05-22
Use new version of issue templates

erikn69:


willpower232 (approved):
Good stuff, I'd probably make the file name lowercase (bug_report.yml instead of Bug_report.yml) unless its required but good to be updated either way.

PR #933, opened 2024-05-22
Avoid only using `id` as audit model primary key

erikn69:
https://github.com/owen-it/laravel-auditing/blob/f236e42c6cf8ba6264499e205d59d4c0b37d1def/config/audit.php#L16

On custom audit implementation, the model primary key could be different to `id`, like `uuid`

---

Also `threshold prune` brings all the audits to memory before prune
I have my doubts `PHP_INT_MAX` 😕

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -77,7 +77,7 @@ trait Audit
 
         // Metadata
         $this->data = [
-            'audit_id'         => $this->id,
+            'audit_id'         => $this->getKey(),
             'audit_event'      => $this->event,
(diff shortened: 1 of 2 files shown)


willpower232 (commented):
I'm a big fan of this however the way you've rewritten the prune makes it a straight database query rather than getting them all into memory first so is the limit actually required?

willpower232 (approved):
Oh I'm blind, yeah that makes sense with the return haha

parallels999 (approved):

PR #934, opened 2024-05-22
12.x EOL

erikn69:
[12.x](https://github.com/owen-it/laravel-auditing/tree/v12) branch has not received changes since [2 years ago](https://github.com/owen-it/laravel-auditing/commit/98f1cfddbc4ed257e5644fe02e97db5674c7571a), and Laravel 9.x  has already reached [end of life](https://laravel.com/docs/9.x/releases#support-policy)
I think the "manpower" should be prioritized for still supported versions

---

On a separate matter I have no knowledge of what was being worked on in [v14-dev](https://github.com/owen-it/laravel-auditing/tree/v14-dev) branch, 
It is quite outdated, and I don't know if all the features were finished

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Thank you for choosing OwenIt\LaravelAuditing!
 Version   | Illuminate     | Status                  | PHP Version
 :----------|:---------------|:------------------------|:------------
 13.x      | 7.x.x - 11.x.x | Active support :rocket: | > = 7.3 \| 8.0
-12.x      | 6.x.x - 9.x.x | Active support          | > = 7.3 \| 8.0
+12.x      | 6.x.x - 9.x.x | End of life             | > = 7.3 \| 8.0
 11.x      | 5.8.x - 8.x.x | End of life             | > = 7.3
 10.x      | 5.8.x - 7.x.x | End of life             | > = 7.2.5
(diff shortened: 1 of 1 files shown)


willpower232 (approved):
Big agree with this, definitely ready to leave that support behind.

Re v14, it looks like mostly refactoring for using newer PHP features but probably quite old at this point. It could be worth doing one PR with [ECS](https://github.com/easy-coding-standard/easy-coding-standard) to fix up all the linting and then repeating the phpstan parts separately?

erikn69:
first it would be better to synchronize v13 and v14-dev, I think that all the linting and then repeating the phpstan can be done just before release, I am worried that some functionality will be lost

Also [v14 pendings](https://github.com/owen-it/laravel-auditing/issues?q=is%3Aissue+is%3Aopen+label%3Av14)

willpower232:
sounds like a plan :pray:

parallels999:
>it would be better to synchronize v13 and v14-dev, I think that all the linting and then repeating the phpstan can be done just before release
>>sounds like a plan 🙏

@willpower232 when you have time, please sync those branches, it would be great if v14 were released at least in beta

angeljqv:
It would be good to release v14, so we could leave PHP 7, Laravel 7-9 behind

PR #938, opened 2024-05-27
Globally disable auditing on `withoutAuditing()`

erikn69:
Discussed on  [#917](https://github.com/owen-it/laravel-auditing/pull/917#issuecomment-2133701358)
```php

User::withoutAuditing(function () use ($user, $inputs, $additional) {
    // no audit
    $user->update($inputs);

    // this will not be audited if the Profile class is Auditable
    $user->profile()->updateOrCreate([], $additional);
}, true); // adding true($globally) disables auditing on entire code block
```
**CC:** @derekmd

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -35,6 +35,13 @@ trait Audit
     protected $modified = [];
 
     /**
+     * Is globally auditing disabled?
+     *
+     * @var bool
+     */
+    public static $auditingDisabled = false;
+
+    /**
      * {@inheritdoc}
      */
     public function auditable()
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -256,7 +256,7 @@ trait Auditable
      */
     public function readyForAuditing(): bool
     {
-        if (static::$auditingDisabled) {
+        if (static::$auditingDisabled || Audit::$auditingDisabled) {
             return false;
         }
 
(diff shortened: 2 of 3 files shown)


willpower232 (commented):
Is it worth converting the other uses of `::$auditingDisabled` to `::isAuditingDisabled()` to promote the getter a bit more?

erikn69:
>Is it worth converting the other uses of `::$auditingDisabled` to `::isAuditingDisabled()` to promote the getter a bit more?

maybe v14, In this case it would be a breaking change, because there is a contract for Auditable(_That's what I understand_)
https://github.com/owen-it/laravel-auditing/blob/master/src/Contracts/Auditable.php

willpower232:
I mostly meant in the tests but I get your point. Definitely worth adding to the interface at some point.

erikn69:
>I mostly meant in the tests

In this case it is possible, but they are only tests and will continue to work, I prefer that the PR be easy to review, that could be changed in a maintenance PR

willpower232 (approved):
Looks good to me then. Do you have permissions on the (not archived) docs repo? Worth amending the PR in there to include the global bit

erikn69:
>Do you have permissions on the (not archived) docs repo?

No, i don't

anteriovieira:
Hi @erikn69 , now you have 🚀

PR #940, opened 2024-05-28
Fix php 8.3 deprecation warning

erikn69:
```
Accessing static trait property OwenIt\Auditing\Audit::$auditingDisabled is deprecated,
it should only be accessed on a class using the trait
```

### DIFF
--- a/src/Models/Audit.php
+++ b/src/Models/Audit.php
@@ -22,6 +22,13 @@ class Audit extends Model implements \OwenIt\Auditing\Contracts\Audit
     protected $guarded = [];
 
     /**
+     * Is globally auditing disabled?
+     *
+     * @var bool
+     */
+    public static $auditingDisabled = false;
+
+    /**
      * {@inheritdoc}
      */
     protected $casts = [
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -35,13 +35,6 @@ trait Audit
     protected $modified = [];
 
     /**
-     * Is globally auditing disabled?
-     *
-     * @var bool
-     */
-    public static $auditingDisabled = false;
-
-    /**
      * {@inheritdoc}
      */
     public function auditable()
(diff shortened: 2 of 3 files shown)


parallels999 (approved):

willpower232 (approved):

src/Models/Audit.php
```diff
@@ -21,6 +21,13 @@ class Audit extends Model implements \OwenIt\Auditing\Contracts\Audit
      */
     protected $guarded = [];
 
+    /**
+     * Is globally auditing disabled?
+     *
+     * @var bool
+     */
+    public static $auditingDisabled = false;
```
Levivb, line 29: What do you think about changing the property name to `$disableAuditingGlobally`? In version `13.6.6` our `Audit` entity broke because it's both using the `Auditable` trait (through a parent class) and the `Audit` trait. We use `public static $auditingDisabled = true;` from the `Auditable` trait to disable the Auditing of the `Audit` entity.

Now, that is fixed in this pr as well, but to make the code more readable renaming this property to `$disableAuditingGlobally` shows its intent more clearly imo

willpower232: No offence but it sounds like your model inheritance is a bit weird, perhaps a cleaner and safer fix would be if the library detected that the model was both an instanceof Auditable and an instanceof Audit and avoided creating some kind of infinite loop on its own :sweat_smile:

Levivb: It was merely a suggestion to rename the property for globally disabling auditing to `$disableAuditingGlobally` for clarity. Regardless of our inheritance tree.

Levivb: Although a check in the library to prevent infinite loops might be a nice addition as well :D 
Not immediately necessary for us since we fixed that on our side :)

Issue #946, opened 2024-06-07
Prepared statement contains too many placeholders


rebbieboi:
### PHP Version

8.1.28

### Laravel Version

10.39.0

### Package Version

13.6.4

### Description

Audit is triggering a delete and in doing so logs with the error below. Yes, there is a lot of IDs sent to it for the IN condition

```
Prepared statement contains too many placeholders
(Connection: mysql, SQL: delete from `audits` where `audits`.`auditable_type` = App\Models\TheModel and `audits`.`auditable_id` = 4 and `audits`.`auditable_id` is not null and `id` in (

#0 /vendor/laravel/framework/src/Illuminate/Database/Connection.php(776): Illuminate\\Database\\Connection->runQueryCallback()
#1 /vendor/laravel/framework/src/Illuminate/Database/Connection.php(612): Illuminate\\Database\\Connection->run()
#2 /vendor/laravel/framework/src/Illuminate/Database/Connection.php(557): Illuminate\\Database\\Connection->affectingStatement()
#3 /vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php(3617): Illuminate\\Database\\Connection->delete()
#4 /vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(1271): Illuminate\\Database\\Query\\Builder->delete()
#5 /vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php(23): Illuminate\\Database\\Eloquent\\Builder->delete()
#6 /vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php(52): Illuminate\\Database\\Eloquent\\Relations\\Relation->forwardCallTo()
#7 /vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Relation.php(517): Illuminate\\Database\\Eloquent\\Relations\\Relation->forwardDecoratedCallTo()
#8 /vendor/owen-it/laravel-auditing/src/Drivers/Database.php(37): Illuminate\\Database\\Eloquent\\Relations\\Relation->__call()
#9 /vendor/owen-it/laravel-auditing/src/Auditor.php(88): OwenIt\\Auditing\\Drivers\\Database->prune()
#10 /vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(355): OwenIt\\Auditing\\Auditor->execute()
[...]

PR #948, opened 2024-06-12
Fix database prune, offset not working on some dbs

### LINKED ISSUE #947: Audited logs deleted on fresh install with threshold set to > 0
### PHP Version

8.3

### Laravel Version

11.x

### Package Version

13.6.7

### Description

Hi there, firstly thanks for the great package!

I'm not sure if this is something I'm doing or what, but I've just installed a fresh copy, followed the instructions to the letter, and it was refusing to keep audit log entries when I'd tried saving in the following manners:

- Via "save" buttons in Filament PHP
- Via tinker `->save()` method on a model instance (with `console` set to `true` in the config)

I'd stopped short of making a custom controller just to test this, as I had added the following to my model:

```php
/**
 * The "booted" method of the model.
 */
protected static function booted(): void
{
    static::updated(function (Property $property) {
        dump('updated', $property);
    });
}
```

And that dump was coming through on the above save methods.

I'd also added an event listener for `OwenIt\Auditing\Events\Auditing` which was dumping out each time I tried the above save methods as well.

It wasn't until I'd set the `threshold` option back to `0` (from my preference of `100`) that it started actually logging anything.

I've tested this with the threshold at various values and it seems that the prune method might have a bug in it as when the value is anything other than `0`, the entire log for that model (and only that model instance, it seems) is removed on each audit event.

### Steps To Reproduce

- Install auditing tool
- Configure as normal
- Create at least one auditable model
- Set threshold to anything other than 0
- Trigger an update/create etc event

### Possible Solutions

_No response_

erikn69:
- Closes #947

### DIFF
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -23,9 +23,13 @@ class Database implements AuditDriver
     public function prune(Auditable $model): bool
     {
         if (($threshold = $model->getAuditThreshold()) > 0) {
+            $class = get_class($model->audits()->getModel());
+            $keyName = (new $class)->getKeyName();
+
             return $model->audits()
-                ->latest()
-                ->offset($threshold)->limit(PHP_INT_MAX)
+                ->whereNotIn(
+                    $keyName, $model->audits()->select($keyName)->limit($threshold)->latest()
+                )
                 ->delete() > 0;
         }
 
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -25,11 +25,15 @@ class Database implements AuditDriver
         if (($threshold = $model->getAuditThreshold()) > 0) {
             $class = get_class($model->audits()->getModel());
             $keyName = (new $class)->getKeyName();
-
+        
+            $idsToKeep = $model->audits()
+                ->select($keyName)
+                ->limit($threshold)
+                ->latest()
+                ->pluck($keyName);
+        
             return $model->audits()
-                ->whereNotIn(
-                    $keyName, $model->audits()->select($keyName)->limit($threshold)->latest()
-                )
+                ->whereNotIn($keyName, $idsToKeep)
                 ->delete() > 0;
         }
 
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -23,17 +23,22 @@ class Database implements AuditDriver
     public function prune(Auditable $model): bool
     {
         if (($threshold = $model->getAuditThreshold()) > 0) {
-            $class = get_class($model->audits()->getModel());
-            $keyName = (new $class)->getKeyName();
-        
-            $idsToKeep = $model->audits()
-                ->select($keyName)
-                ->limit($threshold)
-                ->latest()
-                ->pluck($keyName);
-        
+            $auditClass = get_class($model->audits()->getModel());
+            $auditModel = new $auditClass;
+
             return $model->audits()
-                ->whereNotIn($keyName, $idsToKeep)
+                ->leftJoinSub(
+                    $model->audits()->select($auditModel->getKeyName())->limit($threshold)->latest(),
+                    'audit_threshold',
+                    function ($join) use ($auditModel) {
+                        $join->on(
+                            $auditModel->gettable().'.'.$auditModel->getKeyName(),
+                            '=',
+                            'audit_threshold.'.$auditModel->getKeyName()
+                        );
+                    }
+                )
+                ->whereNull('audit_threshold.'.$auditModel->getKeyName())
                 ->delete() > 0;
         }
 
--- a/tests/Functional/AuditingTest.php
+++ b/tests/Functional/AuditingTest.php
@@ -297,16 +297,23 @@ class AuditingTest extends AuditingTestCase
         ]);
 
         $article = factory(Article::class)->create([
-            'reviewed' => 1,
+            'title' => 'Title #0',
         ]);
 
-        foreach (range(0, 99) as $count) {
+        foreach (range(1, 20) as $count) {
+            if ($count === 11) {
+                sleep(1);
+            }
+
             $article->update([
-                'reviewed' => ($count % 2),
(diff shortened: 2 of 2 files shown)


willpower232 (approved):
I was going to wait for @neil-cldagency to confirm his problem was resolved before reviewing but yeah, this is a great fix :ok_hand: 

Is it worth changing the tests to use mysql rather than sqlite? Or running both somehow?

neil-cldagency:
Hi all — apologies, been deep in development. Will look at this tomorrow AM (BST).

parallels999:
> Is it worth changing the tests to use mysql rather than sqlite? Or running both somehow?

It would be a good idea, I have seen other packages test various DRIVERs
[spatie/laravel-permission/.github/workflows/test-cache-drivers.yml](https://github.com/spatie/laravel-permission/blob/main/.github/workflows/test-cache-drivers.yml)

neil-cldagency:
Getting a consistently occurring error when I attempt to edit any models while the threshold is > 0. Looks like you're using a query not all engines support. Fair enough if it's MySQL only but I presume the package should support at least 5.7+?

Will defer to @parallels999 on that one!

```
SQLSTATE[42000]: Syntax error or access violation: 1235 This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery' 

delete from
  `audits`
where
  `audits`.`auditable_type` = App \ Models \ Property
  and `audits`.`auditable_id` = 1
  and `audits`.`auditable_id` is not null
  and `id` not in (
    select
      `id`
    from
      `audits`
    where
      `audits`.`auditable_type` = App \ Models \ Property
      and `audits`.`auditable_id` = 1
      and `audits`.`auditable_id` is not null
    order by
      `created_at` desc
    limit
      2
  )
```

MySQL version is 8.0.33-0ubuntu0.20.04.2 (supplied via ddev)

willpower232 (dismissed):
I didn't realise I already had a suitable test environment for this problem and I'm now sad that current MySQL versions don't support the elegance of this fix.

I think the query [could be re-written as a join](https://dba.stackexchange.com/a/20222) but writing one manually in Laravel makes my head hurt :sweat_smile: 

The other alternative would be getting the IDs first which seems to work in a pinch.

```php
    $class = get_class($model->audits()->getModel());
    $keyName = (new $class)->getKeyName();

    $keys = $model->audits()
        ->select($keyName)
        ->limit($threshold)
        ->latest()
        ->pluck($keyName);

    return $model->audits()
        ->whereNotIn(
            $keyName,
            $keys
        )
        ->delete() > 0;
}
```

I don't prune audits so don't have a bias any particular way.

erikn69:
Ok, i will reverse it for now, I think it would be the best since I cannot test old versions of the db at the moment

erikn69:
Ok, i will reverse it for now, I think it would be the best since I cannot test old versions of the db at the moment
But the problem would return in case of a high threshold, lease #946

erikn69:
>i think the query [could be re-written as a join](https://dba.stackexchange.com/a/20222)

I'll give it a try

neil-cldagency:
> The other alternative would be getting the IDs first which seems to work in a pinch.

From memory, Laravel itself makes use of "get IDs first then `in`" SQL patterns, so unless there's a realistic chance of exceeding query length limits I wouldn't have anything against using that pattern in this case personally — depends on that risk though, I'd have thought.

erikn69:
@willpower232, @neil-cldagency hi, could you test "join" version?

willpower232:
Looking good to me, it behaves as I think it should in my weird setup. With a threshold of 1 the SQL looks like this and looks like it includes all the correct wheres

```sql
delete `audits`
from `audits`
left join (
	select `id`
	from `audits`
	where `audits`.`auditable_type` = "App\\Models\\User"
	and `audits`.`auditable_id` = 3
	and `audits`.`auditable_id` is not null
	order by `created_at` desc limit 1
) as `audit_threshold` on `audits`.`id` = `audit_threshold`.`id`
where `audits`.`auditable_type` = "App\\Models\\User"
and `audits`.`auditable_id` = 3
and `audits`.`auditable_id` is not null
and `audit_threshold`.`id` is null
```

giolaza:
this fix works, but not sure using left join is good idea for optimization

giolaza (approved):
these fixes work

willpower232:
@neil-cldagency do you have time to test the fix or should we ship based on what else we've seen?

willpower232:
@giolaza as you can read above, the left join is the best option, the most elegant solution isn't supported by mysql, and plucking out the IDs would eventually cause a problem for someone

giolaza:
@willpower232 plucking IDs caused issues when you used it in a subquery with a limit. Maybe a better solution is using Laravel's ->pluck('id').

Before the update, you were using pluck to get the IDs to delete. If we analyze the situation, I think this is the best solution.

Situations:

From model creation/package installation, the limit was more than 0. This means that the number of rows to delete should be small.
If it was unlimited and then became limited, the first log removal will cause a load, then it goes back to the first argument.

parallels999:
>maybe a better solution is using Laravel's `->pluck('id')`

@giolaza originally used `->pluck('id')`(_v13.6.5_), but apparently that caused https://github.com/owen-it/laravel-auditing/issues/946
https://github.com/owen-it/laravel-auditing/blob/27e1b3121a6b34f9635d2dbe84c13ca11ed2bee0/src/Drivers/Database.php#L32

neil-cldagency:
> @neil-cldagency do you have time to test the fix or should we ship based on what else we've seen?

Happy to let others test here @willpower232 as I'm knee deep in an implementation at the moment and testing involves reverting a FilamentPHP plugin each time 😅

willpower232 (approved):
These changes fixed the problem of removing all audits so happy with this

neil-cldagency:
Apologies all, I hadn't checked who regularly maintains this project. Thanks to everyone involved, though!

PR #951, opened 2024-06-26
fix prune method

giolaza:


willpower232 (changes requested):
please see progress in #948

giolaza:
> please see progress in #948

#948  is better because it more dynamic

src/Drivers/Database.php
```diff
@@ -23,10 +23,26 @@ public function audit(Auditable $model): ?Audit
...
+            $class = get_class($model->audits()->getModel());
+            $auditKeyName = (new $class)->getKeyName();
+
+            $forRemoval = $model->audits()
                 ->latest()
-                ->offset($threshold)->limit(PHP_INT_MAX)
-                ->delete() > 0;
+                ->get()
+                ->slice($threshold)
+                ->pluck($auditKeyName);
```
parallels999, line 33: It seems to me that it is more efficient to put the `pluck` above the `slice`, replacing the `get`

src/Drivers/Database.php
```diff
@@ -23,10 +23,26 @@ public function audit(Auditable $model): ?Audit
...
+
+            if (!$forRemoval->isEmpty()) {
+                $forRemovalChunks = $forRemoval->chunk(Config::get('audit.placeholders_limit', 50000));
+                $answer = false;
+                foreach ($forRemovalChunks as $chunk) {
+                    $answer = $answer || $model->audits()
```
parallels999, line 39: Maybe I don't understand this,
after the first `chunk`, `answer` would be `true`,
then in the second chunk(and later), the delete would no longer be executed

parallels999 (commented):
You also need to add tests to corroborate that the chuncks are working.

src/Drivers/Database.php
parallels999: Also, if the first chunk is executed, and the second for some reason is not, would it return `true`/`false` even though only a portion of the records were deleted?

src/Drivers/Database.php
```diff
@@ -23,10 +23,26 @@ public function audit(Auditable $model): ?Audit
...
+            $forRemoval = $model->audits()
                 ->latest()
-                ->offset($threshold)->limit(PHP_INT_MAX)
-                ->delete() > 0;
+                ->get()
+                ->slice($threshold)
+                ->pluck($auditKeyName);
+
+            if (!$forRemoval->isEmpty()) {
+                $forRemovalChunks = $forRemoval->chunk(Config::get('audit.placeholders_limit', 50000));
```
parallels999, line 36: If you have thousands of records, they will all be stored in memory as models, wouldn't that be less efficient than a `leftJoin`?

src/Drivers/Database.php
giolaza: any ideas? I thought if some deletion was  performed in any case return true

src/Drivers/Database.php
```diff
@@ -23,10 +23,26 @@ public function audit(Auditable $model): ?Audit
...
         if (($threshold = $model->getAuditThreshold()) > 0) {
-            return $model->audits()
+            $class = get_class($model->audits()->getModel());
+            $auditKeyName = (new $class)->getKeyName();
+
+            $forRemoval = $model->audits()
                 ->latest()
-                ->offset($threshold)->limit(PHP_INT_MAX)
-                ->delete() > 0;
+                ->skip($threshold)
```
angeljqv, line 31: skip and take are alias of offset and limit

giolaza:
think this topic now is closed

PR #954, opened 2024-07-03
Support for auditSync and auditDetach on Auditable custom Pivot class

txdFrancesco:
Allow the use of auditSync and auditDetach on BelongsToMany relations where the pivot class is an Auditable class.

This is useful in cases where the relation model is itself a Laravel model and may be updated (and thus audited) indipendently  from the related models, without requring dedicated code around every ->auditSync call on the related models.


willpower232 (changes requested):
are you able to correct your spaces to match the rest of the code and add some tests for the functionality you've added?

txdFrancesco:
For the spaces sure, for the tests I'll need to add a couple of models to reproduce. 
Users -> GroupMembers (with 'role' column) -> Groups could be an example of my case. Let me know if you have a better/preferred example structure. I'll update when I'm done, thanks

txdFrancesco:
I added some tests that cover sync, attach and detach with an Auditable custom Pivot class.
AuditAttach is not affected by my code change, I put in the test anyway for completeness. Let me know if it is better to remove it.

The other tests are failing without the code change and are passing now.

src/Auditable.php
```diff
@@ -746,7 +748,16 @@ public function auditDetach(string $relationName, $ids = null, $touch = true, $c
         }
 
         $old = $relationCall->get($columns);
-        $results = $relationCall->detach($ids, $touch);
+        
+        $pivotClass = $relationCall->getPivotClass();
+        
+        if($pivotClass !== Pivot::class && is_a($pivotClass, ContractsAuditable::class, true)) {
```
willpower232, line 754:
```suggestion
        if ($pivotClass !== Pivot::class && is_a($pivotClass, ContractsAuditable::class, true)) {
```

src/Auditable.php
```diff
@@ -772,9 +783,18 @@ public function auditSync(string $relationName, $ids, $detaching = true, $column
         if ($callback instanceof \Closure) {
             $this->applyClosureToRelationship($relationCall, $callback);
         }
-
+        
         $old = $relationCall->get($columns);
```
willpower232, line 787:
```suggestion

        $old = $relationCall->get($columns);
```

src/Auditable.php
```diff
@@ -772,9 +783,18 @@ public function auditSync(string $relationName, $ids, $detaching = true, $column
...
-        $changes = $relationCall->sync($ids, $detaching);
+        
+        $pivotClass = $relationCall->getPivotClass();
+        
+        if($pivotClass !== Pivot::class && is_a($pivotClass, ContractsAuditable::class, true)) {
```
willpower232, line 791:
```suggestion
        if ($pivotClass !== Pivot::class && is_a($pivotClass, ContractsAuditable::class, true)) {
```

willpower232 (changes requested):
some final spaces and maybe wake up github actions

bmoex:
I think `auditAttach` should be checked too if you want to implement fully

txdFrancesco:
> some final spaces and maybe wake up github actions

Sorry for the inconvenience, I'm not sure what should I do to make this pull request go through.
The spacing was fixed by your commit (I just fixed one more line), but it's not clear to me what you mean with "wake up github actions". Could you explain or point me in the right direction please?
Thank you

txdFrancesco:
> I think `auditAttach` should be checked too if you want to implement fully

The bug was caused by the way the detach method creates the query to delete the pivot class (creating a new instance and not retrieving the one in the database). Attaching is not affected by it (there is no instance to retrieve), applying the changes would just remove the "created" audit on the pivot class, which is against the idea of having an auditable pivot class.

A full solution would be to rework how the detach method builds the query instead of wrapping with withoutAuditing, but frankly I don't know the codebase enough to make such an impactful change for what is an edge usecase.

onlime:
> The bug was caused by the way the detach method creates the query to delete the pivot class (creating a new instance and not retrieving the one in the database). Attaching is not affected by it (there is no instance to retrieve), applying the changes would just remove the "created" audit on the pivot class, which is against the idea of having an auditable pivot class.

Thanks a lot @txdFrancesco for this explanation. Would be super nice if this could be supported in the future.

@MortenDHansen is there something in the works for V14? https://github.com/owen-it/laravel-auditing/pull/984/commits/f49ab23ede14fd5c3282264b4a177a22fbb36ae0 looks promising, or at least it's a smart refactor to move that logic into a separate `AuditsPivotRecords` trait.
Thanks a lot for this great package and very cool to see some progress here.

willpower232 (approved):
seems fine to me, not sure when is good to merge this @erikn69 perhaps just needs a pint/phpstan run after merging your other PR?

willpower232 (changes requested):
@txdFrancesco are you able to update the PR now that the factories are modernised?

txdFrancesco (commented):
Should be done, test are passing on my machine. @willpower232 can you approve the action workflow? [here](https://github.com/owen-it/laravel-auditing/actions/runs/13545615910) it says it needs approval from a mantainer

willpower232 (approved):
looking good to me, thanks for the quick response

@erikn69 time for merging and v14?

erikn69:
@willpower232 the tests were not run on this change, I won't have time to review this for at least a couple of weeks, if you reviewed it and it won't cause problems feel free to merge it and make the release for v14, you would also have to put the note for the event breaking change (I couldn't find why it works differently in certain scenarios)

PR #963, opened 2024-08-19
fix: RecordCustomAudit listener

Pr3d4dor:
This pull request fixes this problem:
```
OwenIt\Auditing\Listeners\RecordCustomAudit::handle(): Argument #1 ($model) must be of type OwenIt\Auditing\Contracts\Auditable, OwenIt\Auditing\Events\AuditCustom given, called in /mnt/c/Users/bine/Documents/mapos-laravel/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php on line 478 
```


willpower232:
I'm assuming that fixing these tests also makes sense as well as correcting the listener

Pr3d4dor:
@willpower232 The is a strange behavior, may there is a bug in the Laravel core.

This does not work:
```
Event::dispatch(AuditCustom::class, [$this]);
```

But this works:
```
Event::dispatch(new AuditCustom($this));
```

willpower232 (approved):
These changes make sense to me, I would presume that the current way used to work at some point but only as an accident of how ambiguous `::dispatch` can be.

@erikn69 I think you refactored `src/Auditable.php` back in #866 so presumably it worked for you then but as noted in the linked issue, isn't how laravel intended things to be dispatched so are you happy with these changes?

erikn69:
@willpower232 in #866 just unify the repeated code and small changes, 
if you check it, `Event::dispatch(AuditCustom::class, [$this]);` was always present, and it works in tests

>  When the given "event" is actually an object we will assume it is an event object and use the class as the event name and this event itself as the payload to the handler, which makes object based events quite simple.
[laravel/framework/src/Illuminate/Events/Dispatcher.php#L244-L268](https://github.com/laravel/framework/blob/53091c205f110fb57ffdc1584e10bdaaf10d0d88/src/Illuminate/Events/Dispatcher.php#L244-L268)

The Laravel code specifies that the same event object is sent to the `handler`, only if the first argument is an object, 
That makes me think that something out of the ordinary is happening in that scenario, I can't reproduce it, 

I'm not happy changing the signature, but it is the most feasible solution that I see

zoispag:
This is a breaking change, isn't it?

Many of our tests are now failing with error:

```php
TypeError: OwenIt\Auditing\Listeners\RecordCustomAudit::handle(): Argument #1 ($event) must be of type OwenIt\Auditing\Events\AuditCustom, App\Models\Book given, called in ./vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php on line 479
```

If this is on purpose, a new major version should be released.

PR #967, opened 2024-09-18
allows for merging excludes into the config exclude list

gravitymedianet:
The auditExcludeMerge property overrides the auditExclude property if set, and will be merged into the config's audit.exclude property.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -143,6 +143,14 @@ trait Auditable
      */
     public function getAuditExclude(): array
     {
+        if(property_exists($this, 'auditExcludeMerge'))
+        {
+            return array_merge(
+                $this->auditExcludeMerge,
+                Config::get('audit.exclude', [])
+            );
+        }
+
         return $this->auditExclude ?? Config::get('audit.exclude', []);
     }
 
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -143,12 +143,8 @@ trait Auditable
      */
     public function getAuditExclude(): array
     {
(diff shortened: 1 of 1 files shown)


src/Auditable.php
```diff
@@ -143,6 +143,14 @@ protected function resolveAuditExclusions()
      */
     public function getAuditExclude(): array
     {
+        if(property_exists($this, 'auditExcludeMerge'))
+        {
+            return array_merge(
+                $this->auditExcludeMerge,
+                Config::get('audit.exclude', [])
+            );
+        }
```
parallels999, line 152: what about `bool $auditExcludeMerge `?
```php
if ($this->auditExcludeMerge ?? false) {
    return array_merge($this->auditExclude ?? [], Config::get('audit.exclude', []));
}
```
It seems excessive to me to have two properties for almost the same thing.
Also take a look at [PHP Coding Standards](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer)

gravitymedianet: I like it!

src/Auditable.php
```diff
@@ -143,6 +143,10 @@ protected function resolveAuditExclusions()
      */
     public function getAuditExclude(): array
     {
+        if ($this->auditExcludeMerge ?? false) {
```
parallels999, line 146: Maybe a global option, so you could avoid adding this to each model
```php
if ($this->auditExcludeMerge ?? Config::get('audit.exclude_merge')) {
```

PR #968, opened 2024-09-18
Gravitymedianet audit exclude merge feature

gravitymedianet:
Allows for enabling global merging of the config exclude values and any local auditable model's auditExclude values, instead of replacing the global exclude list.


gravitymedianet:
I closed this and did a new pull request with the global updates

PR #969, opened 2024-09-18
auditExclude merge versus replace

gravitymedianet:
Allows for merging the local auditExclude values into the global exclude list instead of replacing it.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -143,7 +143,10 @@ trait Auditable
      */
     public function getAuditExclude(): array
     {
-        if ($this->auditExcludeMerge ?? false) {
+        if ( 
+            Config::get('audit.exclude_merge', false) || 
+            ($this->auditExcludeMerge ?? false) 
+        ) {
             return array_merge($this->auditExclude ?? [], Config::get('audit.exclude', []));
         }
 
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -143,6 +143,14 @@ trait Auditable
      */
     public function getAuditExclude(): array
     {
+        if(property_exists($this, 'auditExcludeMerge'))
+        {
+            return array_merge(
+                $this->auditExcludeMerge,
+                Config::get('audit.exclude', [])
+            );
+        }
+
         return $this->auditExclude ?? Config::get('audit.exclude', []);
     }
 
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -143,12 +143,8 @@ trait Auditable
      */
     public function getAuditExclude(): array
     {
-        if(property_exists($this, 'auditExcludeMerge'))
-        {
-            return array_merge(
-                $this->auditExcludeMerge,
-                Config::get('audit.exclude', [])
-            );
+        if ($this->auditExcludeMerge ?? false) {
+            return array_merge($this->auditExclude ?? [], Config::get('audit.exclude', []));
         }
 
         return $this->auditExclude ?? Config::get('audit.exclude', []);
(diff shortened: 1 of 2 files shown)


src/Auditable.php
```diff
@@ -143,6 +143,13 @@ protected function resolveAuditExclusions()
      */
     public function getAuditExclude(): array
     {
+        if ( 
+            Config::get('audit.exclude_merge', false) || 
+            ($this->auditExcludeMerge ?? false) 
+        ) {
```
parallels999, line 149: https://github.com/owen-it/laravel-auditing/pull/967#discussion_r1765671881
It seems that you decided not to follow the instructions
```php
if ($this->auditExcludeMerge ?? Config::get('audit.exclude_merge')) {
```
The specific of the model must replace the global, 
IMO, if i set `$auditExcludeMerge = false;`, It is expected that the merge will not be done even though the global `'audit.exclude_merge' => true`

gravitymedianet: I do not see how the code is different, but I updated it. Functionally it seems to be the same. But I updated it.

parallels999: Not the same, 
enjoy: [https://onlinephp.io?s=s7EvyCj](https://onlinephp.io?s=s7EvyCjg5dLXV3DNK0ktUqjMLy1SSM5PSVXISC1K1VFIzcvKr1Tk5VJJz8lPSsxxLE3JLHGtSM4pTUn1TS1KT1WwVSgpKk21BqpIxCKXlphTDJIsSyyKTynNLdDAosreXgGX6ZpArSBIpu15pTk5NLUc5r1BaDs0XMi2HQA%2C&v=8.2.20)

gravitymedianet: Perhaps this:
```
 if (
            (Config::get('audit.exclude_merge', false) && ($this->auditExcludeMerge ?? false) !== false))
        ) {
            return merge_array($this->auditExclude ?? [], Config::get('audit.exclude', []);
        }
```

src/Auditable.php
```diff
@@ -144,8 +144,7 @@ protected function resolveAuditExclusions()
     public function getAuditExclude(): array
     {
         if ( 
-            Config::get('audit.exclude_merge', false) || 
-            ($this->auditExcludeMerge ?? false) 
+            $this->auditExcludeMerge ?? Config::get('audit.exclude_merge', false)
```
parallels999, line 147: >It seems that you decided not to follow the instructions

the `false` is not necessary, the default is `null` and the null on a `if` is already `false`
```php
if ($this->auditExcludeMerge ?? Config::get('audit.exclude_merge')) {
```

gravitymedianet: Apologies, now I understand. Code updated.

parallels999: >Code updated. 

the 'false' is still there

willpower232 (changes requested):
as mentioned by @parallels999, the `, false)` is not necessary

it would also be good to see a few tests for the different combinations of the auditExcludeMerge property and the config value

parallels999:
https://github.com/owen-it/laravel-auditing/pull/969#discussion_r1765814348

>enjoy: [https://onlinephp.io?s=s7EvyCj](https://onlinephp.io?s=s7EvyCjg5dLXV3DNK0ktUqjMLy1SSM5PSVXISC1K1VFIzcvKr1Tk5VJJz8lPSsxxLE3JLHGtSM4pTUn1TS1KT1WwVSgpKk21BqpIxCKXlphTDJIsSyyKTynNLdDAosreXgGX6ZpArSBIpu15pTk5NLUc5r1BaDs0XMi2HQA%2C&v=8.2.20)

I did some basic tests

PR #973, opened 2024-10-13
Phpstan lvl 1

rajmundtoth0:
Hi there,

I was wondering how big a task would be to use PHPstan on this package, so I started to resolve the level 1 issues.

I had to change a few things here and there, hope you like it!

Changes:
 - Factories: changed to the modern version.
 - PHPUnit had some configurations prevented to generate code coverage
 - Removed the `Assert` helper method and used `assertSame` instead of  `Assert()::assertArraySubset`
 - Typehinted a few methods
 - Added PHPStan to the CI

Let me know what you think, I am happy to raise the PHPStan level further.


parallels999:
this would make it more difficult to merge V13 with v14

I think there is no one who reviews v14, perhaps it would be better for this PR to be redirected to V14 and complete what is missing

willpower232:
phpstan now at level 8 so all good thanks 🙏

PR #975, opened 2024-10-14
PHP  Code Standard

rajmundtoth0:
I have extracted the PHP CS part from the `v14-dev` branch, as this one can be included in v13, the biggest changes are the strict options in some cases.


willpower232:
Thanks for the effort, sounds like its annoyed about the risky setting. Is there a need to have a separate script for ci in this case?

rajmundtoth0:
I prefer to have the CI to just display the issues, instead of fixing them. That could prevent unwanted changes. I will add the risky option in a sec.

willpower232:
Ah yes I see now, definitely agree with you on that. I don't understand the risky setting but sounds like it would like you to make even more changes :sweat_smile:

rajmundtoth0:
In composer  I set it to check all folders, while in the original config file, it is looking in only to `/src`. Now both checks all folders. Hopefully this makes it happy - also formatted the files outside of  `/src`

rajmundtoth0:
https://cs.symfony.com/doc/ruleSets/PhpCsFixerRisky.html

parallels999:
If you merge this, also merge v13 and v14 so that they are synchronized

rajmundtoth0:
Fixed

willpower232:
pint is now installed and automatic so thanks for your help but linting is now covered

PR #976, opened 2024-10-15
PHPStal lvl 0 baseline

rajmundtoth0:
This can be merged to `v13` as it is only generates a baseline, without actual code change to ensure no more PHPStan level 0 issues are introduced, such as:
basic checks, unknown classes, unknown functions, unknown methods called on $this, wrong number of arguments passed to those methods and functions, always undefined variables

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,23 @@
+includes:
+    - ./phpstan-baseline.neon
+    - vendor/larastan/larastan/extension.neon
+    - vendor/phpstan/phpstan-mockery/extension.neon
+    - vendor/phpstan/phpstan-phpunit/extension.neon
+parameters:
+    level: 0
+    paths:
+        - ./config
+        - ./database
+        - ./src
+        - ./tests
+    exceptions:
+        implicitThrows: false
+        check:
+            tooWideThrowType: true
+    excludePaths:
+        - tests/database/migrations
+services:
+    -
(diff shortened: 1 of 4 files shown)


phpstan.neon
```diff
@@ -0,0 +1,23 @@
...
+    - vendor/larastan/larastan/extension.neon
+    - vendor/phpstan/phpstan-mockery/extension.neon
+    - vendor/phpstan/phpstan-phpunit/extension.neon
+parameters:
+    level: 0
+    paths:
+        - ./config
+        - ./database
+        - ./src
+        - ./tests
```
erikn69, line 12: In my opinion I don't see the need for tests folder

parallels999:
I think it's better to leave this for v14

rajmundtoth0:
Up to you, but the v14 hasn't changed for 10 months. I think it's better to chunk it up, move whatever is possible to v13, so things are moving. Also, easier to review smaller PRs

phpstan.neon
rajmundtoth0: I rather analyse everything possible, some type mismatch or other issue can generate incorrect test results.

willpower232:
IMHO it is better to not have a baseline, I would rather just fix the issues and that shouldn't be a problem in a project of this size.

I have some time towards the weekend so was planning to make some progress on v14 anyway.

willpower232:
phpstan now at level 8 so all good thanks 🙏

PR #986, opened 2024-12-24
🚨 audit should not have a factory

willpower232:
The whole point of this library is to make audit entries and they are generated via the observer during tests.

If audits are made via factory then it won't necessarily behave like real life because of factory randomness, it is easier to avoid doing that entirely.


erikn69 (approved):
It makes sense, great 👍

PR #992, opened 2025-01-13
Update audits.stub

MahammadBaghirzada:
enhanced auditable_id to be string instead of int


willpower232 (changes requested):
If you need it to be a string for your application then you can publish the migration and edit as you need however for the vast majority of users, `unsignedBigInteger` is correct to match the default of `$table->id()` in regular migrations.

parallels999:
Maybe just a comment like `string for UUID`

PR #994, opened 2025-01-18
♻️ doc types first pass

willpower232:
A first pass of doc types, referencing the v14 branch as well as phpstan level 8, removing duplicated types and adding missing types.

I'm hoping this is basic enough to not cause any issues.

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -307,10 +307,10 @@ trait Audit
     /**
      * Get the Audit tags as an array.
      *
-     * @return array<string>|false
+     * @return array<string>
      */
-    public function getTags()
+    public function getTags(): array
     {
-        return preg_split('/,/', $this->tags, -1, PREG_SPLIT_NO_EMPTY);
+        return preg_split('/,/', $this->tags, -1, PREG_SPLIT_NO_EMPTY)?: [];
     }
 }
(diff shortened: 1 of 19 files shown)


willpower232:
spoiler alert php 7 doesn't support this so would have to wait for v14 I guess

image

src/Audit.php
```diff
@@ -291,9 +292,9 @@ public function getModified(bool $json = false, int $options = 0, int $depth = 5
     /**
      * Get the Audit tags as an array.
      *
-     * @return array
+     * @return array<string>|false
      */
-    public function getTags(): array
+    public function getTags()
```
parallels999, line 297: Why not `?:`
```php
return preg_split('/,/', $this->tags, -1, PREG_SPLIT_NO_EMPTY) ?: [];
```

willpower232: personally I would rather allow the user to detect failure since there could be a big problem with whatever the user is trying to do and covering it up with an empty array would effectively silence any issue unless the user is paying close attention to their output.

could always check for the false and throw an exception I guess which would be a bit clearer to the user than the type error which would probably happen currently.

willpower232: Oh yeah absolutely, phpstan is quite brutal. I've just got a lot of years of not trusting anyone haha.

I'll simplify it :pray:

PR #995, opened 2025-01-18
♻️ deal with null and false returns

willpower232:
Also from phpstan, there are some places where the code ignores potential null and false returns.

In v14, past you went with another level of ternary, I adapted it into more intentional if statements.

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -169,7 +169,13 @@ trait Audit
         }
 
         if (preg_match('/^(\d{4})-(\d{1,2})-(\d{1,2})$/', $value)) {
-            return Date::instance(Carbon::createFromFormat('Y-m-d', $value, Date::now('UTC')->getTimezone())->startOfDay());
+            $date = Carbon::createFromFormat('Y-m-d', $value, Date::now('UTC')->getTimezone());
+
+            if ($date === null) {
+                return $value;
+            }
+
+            return Date::instance($date->startOfDay());
         }
 
         if (preg_match('/^(\d{4})-(\d{2})-(\d{2}) (\d{2}):(\d{2}):(\d{2})$/', $value)) {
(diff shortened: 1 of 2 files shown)


src/Audit.php
```diff
@@ -285,7 +299,15 @@ public function getModified(bool $json = false, int $options = 0, int $depth = 5
             }
         }
 
-        return $json ? json_encode($modified, $options, $depth) : $modified;
+        if ($json === true) {
+            $modified = json_encode($modified, $options, $depth);
+
+            if ($modified === false) {
```
erikn69, line 305: Here this code is repeated twice, why not just elvis ternary and early return
Something like this
```php
    // it seems more readable to me
    if (! $json) {
        return $data;
    }

    return json_encode($data, $options, $depth) ?: '{}';
```
Also, If it should be a json it should not be `return '{}';` instead of `return [];`
let's avoid if inside if, unnecessarily

willpower232: I'll give it a go and see what stan thinks :pray:

src/Audit.php
```diff
@@ -169,7 +169,13 @@ private function castDatetimeUTC($model, $value)
         }
 
         if (preg_match('/^(\d{4})-(\d{1,2})-(\d{1,2})$/', $value)) {
-            return Date::instance(Carbon::createFromFormat('Y-m-d', $value, Date::now('UTC')->getTimezone())->startOfDay());
+            $date = Carbon::createFromFormat('Y-m-d', $value, Date::now('UTC')->getTimezone());
+
+            if ($date === null) {
```
parallels999, line 174: Isn't it better to just check falsy?? 
```diff
-if ($date === null) {
+if (! $date) {
```
In any case, we already know that it could be null

PR #997, opened 2025-01-18
♻️ float no more

willpower232:
phpstan was annoyed about using a float in the `substr`, past you had cast it to int at the end, I thought it was slightly less messy to cast the result of `ceil` as that is where the float is introduced for some reason known only to the developers of PHP.


parallels999:
https://www.php.net/manual/es/function.ceil.php
>The return value of `ceil()` is still of type **float** as the value range of **float** is usually bigger than that of **int**.

erikn69 (approved):

PR #998, opened 2025-01-23
PHPStan Fixes

erikn69:


src/Events/DispatchAudit.php
```diff
@@ -71,8 +71,13 @@ public function __serialize()
      */
     public function __unserialize(array $values): void
     {
-        $this->model = new $values['class'];
+        $model = new $values['class'];
 
+        if(! $model instanceof Auditable) {
```
willpower232, line 76:
```suggestion
        if (! $model instanceof Auditable) {
```

willpower232:
This is looking good, thanks for fixing them. Definitely covered the ones I couldn't get round my head but apparently now it is sad about the config 

```
 ------ ---------------------------------------------------------------------------------------------------------- 
  Line   config/audit.php                                                                                          
 ------ ---------------------------------------------------------------------------------------------------------- 
  5      Called 'env' outside of the config directory which returns null when the config is cached, use 'config'.  
         🪪  larastan.noEnvCallsOutsideOfConfig                                                                    
 ------ ---------------------------------------------------------------------------------------------------------- 
 ```

Is it worth starting a new branch for v14 which can include larastan 3.0 and drop PHP 7 etc? Then we could have a phpstan.neon config file which this error could be ignored in.

**Comment by willpower232** [comment]
I also wonder if it would be worth locking in to specific versions of phpstan and larastan so that random new errors don't crop up all the time.

File: src/Events/DispatchAudit.php (hunk 1 above)
Comment (erikn69) [reply src/Events/DispatchAudit.php:76]: Pint will take care of this, but this must be merged first

**Comment by erikn69** [comment]
> Called 'env' outside of the config directory

I think I'll fix that by uploading the action

> I also wonder if it would be worth locking in to specific versions of phpstan and larastan so that random new errors don't crop up all the time.

I was thinking of keeping it in action and always using the latest version, what do you think?
I want to avoid increasing the maintenance load

**Comment by willpower232** [comment]
phpstan is really good at adding more checks every new version so if it wasn't locked at a particular version, new PRs would end up triggering errors in code they hadn't changed.

Obviously we can't have a composer.lock file so I think the best way would be having the specific version number without the `^` in composer.json and then it would work exactly the same locally as it does in actions

**Comment by erikn69** [comment]
Whatever, this PR does not add any github action, nor does it change the composer.json

File: src/Resolvers/UrlResolver.php [hunk 2]
```diff
@@ -5,13 +5,14 @@
...
+use OwenIt\Auditing\Contracts\Resolver;
 
-class UrlResolver implements \OwenIt\Auditing\Contracts\Resolver
+class UrlResolver implements Resolver
 {
     public static function resolve(Auditable $auditable): string
     {
         if (! empty($auditable->preloadedResolverData['url'] ?? null)) {
-            return $auditable->preloadedResolverData['url'];
+            return $auditable->preloadedResolverData['url'] ?? null;
```
parallels999, line 15:
```suggestion
            return $auditable->preloadedResolverData['url'] ?? '';
```

parallels999 (approved):

josevv28 (approved):

willpower232 (approved):

PR #999, opened 2025-01-23
Add pint workflow - Auto fix styling

erikn69:
- Alternative to #975

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,7 @@ tests            export-ignore
 .gitattributes   export-ignore
 .github          export-ignore
 .gitignore       export-ignore
+.pint.json       export-ignore
 .scrutinizer.yml export-ignore
 .styleci.yml     export-ignore
 .travis.yml      export-ignore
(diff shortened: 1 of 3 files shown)


parallels999 (approved):

willpower232 (changes requested):
I don't hate linting but what I would rather is that the linting tools were part of composer.json so whoever is working on it can lint before they commit rather than having to pull random lints all the time

erikn69:
> so whoever is working on it can lint before they commit

Many of those who want to contribute will not have the complete knowledge or might skip it, it is better to leave it automatic, if you check the biggest and more popular packages they have it, I want to avoid extra work for developers or something being overlooked without realizing it. 

If you need examples of packages that use them I could mention them.

parallels999:
> having to pull random lints all the time

You could add a custom `pint.json`, but in my opinion it is better to leave it as is

>[Laravel Pint](https://github.com/laravel/pint) is an opinionated PHP code style fixer for minimalists. Pint is built on top of **PHP-CS-Fixer** and makes it simple to ensure that your code style stays clean and consistent.

willpower232 (approved):

PR #1000, opened 2025-01-23
Update .gitignore

erikn69:


willpower232 (approved):
Sure thing, if that is what happens for you then definitely need to ignore it

erikn69:
Since the 10.x change, but since we also support older versions, you have to have both.

PR #1001, opened 2025-01-24
Add Phpstan workflow

erikn69:
For discussion https://github.com/owen-it/laravel-auditing/pull/998#issuecomment-2612586294
depends on #998

https://github.com/erikn69/laravel-auditing/actions/runs/12992516652/job/36232547600

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -16,7 +16,7 @@ jobs:
             - name: Setup PHP
               uses: shivammathur/setup-php@v2
               with:
-                  php-version: 8.3
+                  php-version: 8.4
                   coverage: none
 
             - name: Install composer dependencies
@@ -24,7 +24,7 @@ jobs:
 
             - name: Install larastan
               run: |
-                  composer require "nunomaduro/larastan" --no-interaction --no-update
+                  composer require "larastan/larastan" --no-interaction --no-update
                   composer update --prefer-dist --no-interaction
 
             - name: Run PHPStan
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,31 @@
+name: PHPStan
+
+on:
+    push:
+        paths:
+            - '**.php'
+            - 'phpstan.neon.dist'
+
+jobs:
+    phpstan:
+        name: phpstan
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: 8.3
+                  coverage: none
+
+            - name: Install composer dependencies
+              uses: ramsey/composer-install@v3
+
+            - name: Install larastan
+              run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,31 @@
...
+              with:
+                  php-version: 8.3
+                  coverage: none
+
+            - name: Install composer dependencies
+              uses: ramsey/composer-install@v3
+
+            - name: Install larastan
+              run: |
+                  composer require "nunomaduro/larastan" --no-interaction --no-update
```
willpower232, line 27: they moved it to larastan/larastan

phpstan-baseline.neon
willpower232, line 1: can do without this file if it is empty, don't want to encourage people hiding issues

phpstan.neon.dist
```diff
@@ -0,0 +1,16 @@
+includes:
+    - phpstan-baseline.neon
+    - ./vendor/nunomaduro/larastan/extension.neon
```
willpower232, line 3: if you `composer require phpstan/extension-installer` then you don't need to have an `includes` section

willpower232 (changes requested):

phpstan.neon.dist
parallels999: already checked [phpstan/extension-installer](https://github.com/phpstan/extension-installer)
I don't think it's necessary to add that whole package just for this line.
I think that only with larastan we have enough

I try to keep things as simple as possible.

Also
>Starting from Composer 2.2.0 you'll get the following question:
>```
>phpstan/extension-installer contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
>Do you trust "phpstan/extension-installer" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?]
>```
>Answer with `y`  to allow the plugin.

>It's best to set the extension's composer package [type](https://getcomposer.org/doc/04-schema.md#type) to phpstan-extension for this plugin to be able to recognize it and to be [discoverable on Packagist](https://packagist.org/explore/?type=phpstan-extension).
>Add phpstan key in the extension composer.json's extra section:
>```json
>{
>  "extra": {
>    "phpstan": {
>      "includes": [
>        "extension.neon"
>      ]
>    }
>  }
>}
>```

phpstan-baseline.neon
erikn69: Done

phpstan.neon.dist
parallels999: I think we could do without that plugin, it can start simple and basic.

parallels999: This is a starting point for v13, not everything should necessarily be applied now,
we will have to analyze if another extension is added or if larastan is enough, 
also maybe needs composer.json changes(_but on other PR_)
in my opinion we should skip `phpstan/extension-installer` for now

erikn69: It can be changed in another PR, for now **phpstan/extension-installer** is excessive for a single extension

willpower232 (approved):

PR #1004, opened 2025-02-05
Laravel 12.x Support

erikn69:


willpower232:
I will have some time in a couple of saturdays, I was going to prep a fresh branch with just the removal of old PHP and code and removal of old Laravel from the v14 branch, and addition of Laravel 12 so that v14 could start from Laravel 12s release date or thereabouts. What do you think?

erikn69:
> Laravel 12 will be launched on February 24th
> Laravel 10 reached end of life on February 4th

It seems like a good idea to me, as long as it goes with the release date, also Laravel 10.x reached its end of life yesterday, v14 could be from 11.x onwards

PR #1006, opened 2025-02-15
🔧 drop php < 8.2 and add laravel 12

willpower232:
As discussed in a few other places and referencing the v14-dev branch

- remove mentions of lumen
- upgrade old style factories
- drop everything before laravel 11 and php 8.2
- turns out this also means phpunit 11 is the minimum version so I let phpunit update the config file which may or may not have been a bad idea
- v13-dev hasn't existed for a while so the mention in composer.json seems broken so I removed
- v14-dev also changed the workflow to run on all branches so I got that too, not sure how useful that is and of course they haven't released laravel 12 yet so it won't pass
- the apt update in the workflow dates back to php 7.4 so I removed too
- my editor alphabetised the dependencies in composer.json, I think that is an option somewhere but didn't go looking for it

phpunit 11 is flagging deprecations for all the `@test`, `@dataProvider`, etc because phpunit 12 drops them in favour of attributes but as I see #1003 renames every test function I didn't want to conflict by adding `test_` to each one.

Presumably we could also do promoted constructor properties on the event classes?

Replaces #1004


erikn69:
> phpunit 11 is flagging deprecations for all the @test, @dataProvider, etc because phpunit 12 drops them in favour of attributes but as I see https://github.com/owen-it/laravel-auditing/pull/1003 renames every test function I didn't want to conflict by adding test_ to each one.

Merged

composer.json
```diff
@@ -58,16 +57,11 @@
...
+            "OwenIt\\Auditing\\Tests\\": "tests/",
+            "Database\\Factories\\OwenIt\\Auditing\\Tests\\Models\\": "tests/database/factories"
         }
     },
-    "suggest": {
-        "irazasyed/larasupport": "Needed to publish the package configuration in Lumen"
-    },
     "extra": {
-        "branch-alias": {
-            "dev-master": "v13-dev"
```
parallels999, line 69: keep the alias? or will it no longer exist?
```json
"branch-alias": {
    "dev-master": "master"
```

willpower232:
@erikn69 sorry if we overlapped, I think I'm all done with this now so whatever you need to do is cool :pray:

willpower232:
I was just going over something else and spotted that the registerAuditingServiceProvider function in the install command isn't relevant for new laravel 11+ installs because the app config file doesn't list service providers any more

https://github.com/laravel/laravel/blob/11.x/config/app.php

Should probably just remove the function?

erikn69:
I did a quick review and everything is in order, I don't have any suggestions

> Should probably just remove the function?

has been unnecessary for a long time, It is safe to remove it

erikn69 (approved):

parallels999:
What about https://github.com/owen-it/laravel-auditing/commit/f49ab23ede14fd5c3282264b4a177a22fbb36ae0

aglipanci:
@erikn69 any chance this can be released since the Larave 12 is released :)

parallels999:
I think there are two pending to release v14

- https://github.com/owen-it/laravel-auditing/pull/963#issuecomment-2664887215
- https://github.com/owen-it/laravel-auditing/pull/1006#issuecomment-2663997561

willpower232:
There is definitely some interesting parts left in the v14 branch but I wouldn't have called them blockers on the release.

I've got a note in my reminders to update the documentation for the event dispatch change.

Issue #1019, opened 2025-05-05
mysql AuditDriver - prune experienced slow query issue


zebastian:
### PHP Version

8.2.6

### Laravel Version

10.48

### Package Version

13.6.8

### Description

sometimes really rarely the prune method hangs as the delete takes really really long to finish (almost an hour).

our use case is using laravel-audits as an audit for contact changes, the audits table contains ~200million records of some million contacts.

slow query log excerpt:

SET timestamp=1745938126;
delete `audits` from `audits` left join (select `id` from `audits` where `audits`.`auditable_type` = 'App\\cfab9d1953' and `audits`.`auditable_id` = 3210790 and `audits`.`auditable_id` is not null order by `created_at` desc limit 30) as `audit_threshold` on `audits`.`id` = `audit_threshold`.`id` where `audits`.`auditable_type` = 'App\\cfab9d1953' and `audits`.`auditable_id` = 3210790 and `audits`.`auditable_id` is not null and `audit_threshold`.`id` is null;
# Time: 250429 10:48:50
# User@Host: mydb-live[mydb-live] @  [172.17.12.195]
# Thread_id: 5651235  Schema: mydb-live  QC_hit: No
# Query_time: 2.003077  Lock_time: 0.000026  Rows_sent: 1  Rows_examined: 0
# Rows_affected: 0  Bytes_sent: 80
--
# User@Host: mydb-live[mydb-live] @  [172.17.12.195]
# Thread_id: 5667915  Schema: mydb-live  QC_hit: No
# Query_time: 3103.636623  Lock_time: 0.000022  Rows_sent: 0  Rows_examined: 146805805
# Rows_affected: 1  Bytes_sent: 11
SET timestamp=1745942497;
delete `audits` from `audits` left join (select `id` from `audits` where `audits`.`auditable_type` = 'App\\cfab9d1953' and `audits`.`auditable_id` = 10254368 and `audits`.`auditable_id` is not null order by `created_at` desc limit 30) as `audit_threshold` on `audits`.`id` = `audit_threshold`.`id` where `audits`.`auditable_type` = 'App\\cfab9d1953' and `audits`.`auditable_id` = 10254368 and `audits`.`auditable_id` is not null and `audit_threshold`.`id` is null;
# Time: 250429 12:02:00
# User@Host: mydb-live[mydb-live] @  [172.17.12.195]
# Thread_id: 5651235  Schema: mydb-live  QC_hit: No
[...]

PR #1021, opened 2025-05-06
[v14] Phpstan fixes

erikn69:
https://github.com/owen-it/laravel-auditing/pull/933/files
In the old version (`->get()->slice($threshold)->pluck('id')`) Memory performance was failing, since with `get()` all the data was brought as eloquent models

Now (`->pluck($keyName)->all()`) `pluck/all` is done directly, which would only use an int array

`whereIn` was replaced by `whereIntegerInRaw`(avoid too many Placeholders issue)

The same solution works on older Laravel versions, if merged it would also be released in v13: https://github.com/owen-it/laravel-auditing/actions/runs/14862464393/job/41730649233

UPDATE: `whereIntegerInRaw` would cause problems when the key is a string, like UUID?

### DIFF
--- a/src/Audit.php
+++ b/src/Audit.php
@@ -183,7 +183,13 @@ trait Audit
         }
 
         if (preg_match('/^(\d{4})-(\d{2})-(\d{2}) (\d{2}):(\d{2}):(\d{2})$/', $value)) {
-            return Date::instance(Carbon::createFromFormat('Y-m-d H:i:s', $value, Date::now('UTC')->getTimezone()));
+            $date = Carbon::createFromFormat('Y-m-d H:i:s', $value, Date::now('UTC')->getTimezone());
+
+            if (! $date) {
+                return $value;
+            }
+
+            return Date::instance($date);
         }
 
         try {
--- a/src/Drivers/Database.php
+++ b/src/Drivers/Database.php
@@ -24,21 +24,17 @@ class Database implements AuditDriver
         if (($threshold = $model->getAuditThreshold()) > 0) {
             $auditClass = get_class($model->audits()->getModel());
             $auditModel = new $auditClass;
+            $keyName = $auditModel->getKeyName();
 
             return $model->audits()
                 ->leftJoinSub(
-                    $model->audits()->getQuery()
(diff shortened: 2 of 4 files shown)


erikn69:
> whereIntegerInRaw would cause problems when the key is a string, like UUID?

@willpower232 ??

willpower232:
True, you would have to edit the ID column of the audits table to actually make that a problem though.

Depends how much of a risk you think that is, that someone would both enable pruning AND change the type of the ID column.

erikn69:
would something like this work?

```php
->whereRaw($keyName, "'".implode("', '", $forRemoval)."'")
```

I think in MariaDb, it would cover int and string, but I don't know about other DB managers

erikn69:
https://github.com/laravel/framework/blob/94c26967f8f7eebd61af49a90f1ee49b1bb6758b/src/Illuminate/Database/Query/Grammars/Grammar.php#L389-L393

I see that the same thing is used in all the grammars, but the question was if the field is int, when making it a string `"My_INT"`, does it work the same in any DB manager? PostGreSQL for example

erikn69:
Making some comparisons it seems to me that the current `leftJoinSub` has better performance, so in this PR I will only leave the phpstan fixes.

Also https://github.com/owen-it/laravel-auditing/issues/1019#issuecomment-2855030639
>btw: we deployed today to the live box:
The issue mentioned in this ticket did not happen

In case it is used again in the future, the code that was there is:
```php
public function prune(Auditable $model): bool
{
        $auditClass = get_class($model->audits()->getModel());
        $keyName = (new $auditClass)->getKeyName();
        $forRemoval = array_slice(
            $model->audits()->latest()->pluck($keyName)->all(),
            $threshold
        );
        
        if (count($forRemoval)) {
            if (is_string($forRemoval[0])) {
                $forRemoval = "'" . implode("', '", $forRemoval) . "'";
            } else {
                $forRemoval = implode(', ', $forRemoval);
            }

            return $model->audits()
                ->whereRaw($keyName ." in ($forRemoval)")
                ->delete() > 0;
        }
    }

    return false;
}
```

PR #1023, opened 2025-05-14
Typehint `created_at` and `updated_at` values for phpstan

aSeriousDeveloper:
I've been working on implementing some functionality for managing audit history in Filament, and I noticed that PHPStan gives an error for `created_at` and `updated_at` being undefined properties.

This PR type-hints them alongside some of the other model properties to resolve this.

### DIFF
--- a/src/Models/Audit.php
+++ b/src/Models/Audit.php
@@ -9,6 +9,8 @@ use Illuminate\Database\Eloquent\Model;
  * @property string $event
  * @property array<string,mixed> $new_values
  * @property array<string,mixed> $old_values
+ * @property Carbon|null $created_at
+ * @property Carbon|null $updated_at
  * @property mixed $user
  * @property mixed $auditable.
  */


parallels999:
But those fields are not available if you do not use `Illuminate\Database\Eloquent\SoftDeletes`(_here it doesn't_); and to do it, you have to customize the model, in that case the typehint can be added in that custom model, 

https://github.com/owen-it/laravel-auditing/blob/177715ebfe23d084394182db62a03df23a31c8fc/config/audit.php#L9-L16

aSeriousDeveloper:
@parallels999 aren't those fields independent of whether soft deletion is enabled? that just controls the `deleted_at` column.

The default migration stub already includes the `timestamps()` as it generated the table. This signals to me that these columns are available within the default provided Audit Model.

https://github.com/owen-it/laravel-auditing/blob/177715ebfe23d084394182db62a03df23a31c8fc/database/migrations/audits.stub#L30-L34

parallels999:
You're right, sorry, I got confused, I was doing other things.

erikn69 (approved):

PR #1032, opened 2025-08-18
Added support to specify a different implementation of the queue listener

markieo1:
Hi,

As discussed in  #1031, hereby the PR.

The code defaults to the `ProcessDispatchAudit`. When a different implementation is provided or `null`, Laravel uses this. Laravel already handles the `null` correctly.

Using it this way anyone can provide a custom listener implementation.

### DIFF
--- a/config/audit.php
+++ b/config/audit.php
@@ -183,7 +183,7 @@ return [
         'connection' => 'sync',
         'queue' => 'default',
         'delay' => 0,
-        'listener' => OwenIt\Auditing\Listeners\ProcessDispatchAudit::class
+        'listener' => OwenIt\Auditing\Listeners\ProcessDispatchAudit::class,
     ],
 
     /*


config/audit.php
```diff
@@ -183,6 +183,7 @@
         'connection' => 'sync',
         'queue' => 'default',
         'delay' => 0,
+        'listener' => OwenIt\Auditing\Listeners\ProcessDispatchAudit::class
```
willpower232, line 186:
```suggestion
        'listener' => OwenIt\Auditing\Listeners\ProcessDispatchAudit::class,
```

willpower232 (commented):
LGTM, @erikn69 did you have an opinion on whether we should either allow this to be overridden or just provide the extra option as mentioned in #1031 ?

src/AuditingServiceProvider.php
```diff
@@ -26,7 +27,7 @@ public function boot()
         $this->mergeConfigFrom(__DIR__.'/../config/audit.php', 'audit');
 
         Event::listen(AuditCustom::class, RecordCustomAudit::class);
-        Event::listen(DispatchAudit::class, ProcessDispatchAudit::class);
+        Event::listen(DispatchAudit::class, Config::string('audit.queue.listener', ProcessDispatchAudit::class));
```
erikn69, line 30: @willpower232 hi, maybe an interface check, to prevent someone from setting up a class that is not a listener, or some non-existent class

erikn69: It is a good option, another could be to create a new contract that requires having a `handle` method and validate that interface.

PR #1037, opened 2025-11-17
make auditable on resolver be nullable

BinaryKitten:
### Details

By default the resolver contract requires that an Auditable be passed through, but for most of the time - there's a null coalesce used against the auditable's ->preloadedResolverData. By allowing these to be optional/nullable, this allows the default fallback to be used where you don't have an Auditable instantiated at this time.

#### Changes
- update contract to allow $auditable to be optional/nullable
- update all contract usages to match updated contract
- update all calls in resolvers to use the nullsafe operator (from php 8.0 ) to maintain existing functionality

### DIFF
--- a/src/Contracts/Resolver.php
+++ b/src/Contracts/Resolver.php
@@ -5,5 +5,5 @@ namespace OwenIt\Auditing\Contracts;
 interface Resolver
 {
     /** @return mixed */
-    public static function resolve(Auditable $auditable);
+    public static function resolve(?Auditable $auditable = null);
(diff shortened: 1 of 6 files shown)


parallels999:
It is a breaking change for anyone using the interface to have their custom "Resolver", The fact that you had to modify `TenantResolver` should show you the incompatibility.

>  without the requirement for the $auditable to be present/passed.

What are you trying to audit? Because in this package, in order to create an audit log, must have an auditable component

PR #1040, opened 2025-11-28
Also audit when model is already bootstrapped before auditing was enabled

gisostallenberg:


src/AuditableObserver.php
```diff
@@ -92,6 +92,10 @@ protected function dispatchAudit(Auditable $model): void
         if (! $model->readyForAuditing()) {
             return;
         }
+        $modelClass = $model::class;
+        if (method_exists($modelClass, 'isAuditingEnabled') && ! $modelClass::isAuditingEnabled()) {
```
willpower232: it seems like one check has been replaced by two, is this one definitely needed? should there be a separate test to pick it up?

erikn69: Because `isAuditingEnabled` is not in the contract, it needs to be added.

src/Auditable.php
```diff
@@ -77,9 +77,7 @@ trait Auditable
      */
     public static function bootAuditable()
     {
-        if (App::getFacadeRoot() && static::isAuditingEnabled()) {
```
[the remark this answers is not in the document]
erikn69: `App::getFacadeRoot()` returned after merge

willpower232 (approved):

PR #1047, opened 2026-03-23
Show deprecations on phpunit

erikn69:
@willpower232 I thought about `failOnDeprecation`, but that would imply that the vendor's deprecation would also cause fails. Example: Laravel 11 with PHP 8.5 tests

But at least we can display deprecations messages to decide how to proceed

What do you think about it?


parallels999 (approved):

willpower232 (approved):

willpower232:
oh of course I was using an older version for those deprecated functions :facepalm:

PR #1049, opened 2026-03-26
Reuse isAuditingEnabled() method

erikn69:
minor nitpick


parallels999 (approved):

willpower232 (approved):
ah yes good spot, it is a shame that isAuditingDisabled and isAuditingEnabled are not mirrors of each other but at this is a good tidy up for now

PR #1051, opened 2026-04-03
Add return type hints to Auditable and Auditor contracts

rrodrigofranco:
## What this PR does

Adds native PHP return type hints to methods in `Auditable` and `Auditor` contracts that were either missing them or had incorrect PHPDoc annotations:

- `getAuditEvent(): ?string` — previously had no return type hint
- `getAuditDriver(): string` — previously had no return type hint and incorrect `@return string|null` PHPDoc (the method always returns a string via fallback)
- `execute(): void` — previously had no return type hint

## Why

Improves compatibility with static analysis tools (PHPStan, Psalm) and makes the API contracts more explicit without any breaking changes to existing behavior.

### DIFF
--- a/src/Auditable.php
+++ b/src/Auditable.php
@@ -479,7 +479,7 @@ trait Auditable
     /**
      * {@inheritdoc}
      */
-    public function getAuditEvent()
+    public function getAuditEvent(): ?string
     {
         return $this->auditEvent;
     }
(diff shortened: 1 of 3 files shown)


willpower232:
these could be breaking changes for some peoples usage of the library and the types present in the docblocks should be sufficient for people who do use static analysis tools so I don't think this should be merged

unless @erikn69 disagrees, I presume the reason these types are only in docblock form is specifically for this flexibility

erikn69:
I think the same; this should have been added before the v14 release, but it's too late now.

PR #1054, opened 2026-06-09
Avoid stopping other update events when restoring

### LINKED ISSUE #1053: Updates to other auditable models inside a restored listener are not audited
### PHP Version

8.3.6

### Laravel Version

14.0.3

### Package Version

14.0.3

### Description

When an auditable soft-deleted model is restored, updates to a different auditable model performed inside that model's `restored` listener are not audited.

This appears to happen because `AuditableObserver::$restoring` is a single static/global boolean. It is set to `true` in `AuditableObserver::restoring()` to suppress the duplicate `updated` audit that Laravel emits while restoring the model. However, while that flag is true, `AuditableObserver::updated()` suppresses **all** `updated` audits, including updates for other auditable model instances/classes triggered during the restore callback.

### Steps To Reproduce

Given two auditable models, where `PurchaseItem` uses soft deletes and belongs to an `Item`:

```php
use Illuminate\Database\Eloquent\SoftDeletes;
use OwenIt\Auditing\Auditable;
use OwenIt\Auditing\Contracts\Auditable as AuditableContract;

class Item extends Model implements AuditableContract
{
    use Auditable;

    protected $fillable = ['status'];
}

class PurchaseItem extends Model implements AuditableContract
{
    use Auditable;
    use SoftDeletes;

    public function item()
    {
        return $this->belongsTo(Item::class);
    }
} 
```
Register a `restored` listener that updates the related auditable model:

```php
PurchaseItem::restored(function (PurchaseItem $purchaseItem): void {
    $purchaseItem->item->update(['status' => 'incoming']);
});
```

Then restore a previously deleted `PurchaseItem`:

```php
$purchaseItem->delete();
$auditsAfterDelete = $purchaseItem->item->audits()->count();

$purchaseItem->restore();

$this->assertSame(
    $auditsAfterDelete + 1,
    $purchaseItem->item->audits()->count(),
    'The Item update performed during PurchaseItem::restored should be audited.'
);
```
### Actual behavior

The related `Item` is updated successfully, but no `updated` audit is created for the `Item`.

### Expected behavior

The package should suppress only the duplicate `updated` event for the model currently being restored. Updates to other auditable models triggered during `restored` listeners should still be audited normally.

erikn69:
Closes #1053

### DIFF
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -13,7 +13,7 @@ class AuditableObserver
     /**
      * Is the model being restored?
      *
-     * @var array
+     * @var array<string,bool>
      */
     public static $restoring = [];
 
--- a/src/AuditableObserver.php
+++ b/src/AuditableObserver.php
@@ -13,9 +13,9 @@ class AuditableObserver
     /**
      * Is the model being restored?
      *
-     * @var bool
+     * @var array
      */
-    public static $restoring = false;
+    public static $restoring = [];
 
     /**
      * Handle the retrieved event.
@@ -45,9 +45,11 @@ class AuditableObserver
     public function updated(Auditable $model)
     {
         // Ignore the updated event when restoring
-        if (! static::$restoring) {
-            $this->dispatchAudit($model->setAuditEvent('updated'));
+        if (static::$restoring[get_class($model) . '_' . $model->getKey()] ?? false) {
+            return;
         }
+
+        $this->dispatchAudit($model->setAuditEvent('updated'));
     }
 
     /**
@@ -70,7 +72,7 @@ class AuditableObserver
         // When restoring a model, an updated event is also fired.
         // By keeping track of the main event that took place,
         // we avoid creating a second audit with wrong values
-        static::$restoring = true;
+        static::$restoring[get_class($model) . '_' . $model->getKey()] = true;
     }
 
     /**
@@ -84,7 +86,7 @@ class AuditableObserver
 
         // Once the model is restored, we need to put everything back
         // as before, in case a legitimate update event is fired
-        static::$restoring = false;
+        unset(static::$restoring[get_class($model) . '_' . $model->getKey()]);
     }
 
     protected function dispatchAudit(Auditable $model): void
--- a/.github/workflows/fix-php-code-style-issues.yml
+++ b/.github/workflows/fix-php-code-style-issues.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Fix PHP code style issues
         uses: aglipanci/laravel-pint-action@v2
@@ -21,6 +21,6 @@ jobs:
           configPath: ".pint.json"
 
       - name: Commit changes
-        uses: stefanzweifel/git-auto-commit-action@v6
+        uses: stefanzweifel/git-auto-commit-action@v7
         with:
           commit_message: Fix styling
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -14,7 +14,7 @@ jobs:
         env:
           COMPOSER_NO_INTERACTION: 1
         steps:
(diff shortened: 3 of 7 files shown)


willpower232 (approved):
good stuff, probably worth shipping without waiting for feedback from the issue?

Edmasoud95:
Just tested it and this works now. Thanks!

src/AuditableObserver.php
```diff
@@ -45,9 +45,11 @@ public function created(Auditable $model)
     public function updated(Auditable $model)
     {
         // Ignore the updated event when restoring
-        if (! static::$restoring) {
-            $this->dispatchAudit($model->setAuditEvent('updated'));
+        if (static::$restoring[get_class($model) . '_' . $model->getKey()] ?? false) {
```
kristapscodes, line 48: Heads up — this line breaks models whose primary key is cast to an enum, and it's been shipping since 14.0.4.

Model::getKey() returns $this->getAttribute($this->getKeyName()), so when the key column has a cast it hands back the cast value rather than a scalar. For an enum-cast primary key that's an enum instance, and concatenating it into the array key throws Error: Object of class … could not be converted to string. We hit it on two models that had worked fine for years.

It also can't be worked around through configuration, because the concatenation runs before dispatchAudit() — so even with audit.console = false, a model saved from a console command still throws, since the enabled check further down is never reached. The same concatenation appears in restoring() and restored(), so all three need the same treatment.

The underlying assumption is that getKey() is stringable. Laravel doesn't make that assumption where it needs a scalar from a key — Builder::castBinding() normalises it via enum_value() before use.

Would you consider normalising the key before building the cache key? Something like:

```php
protected function restoringKey(Auditable $model): string
{
    $key = $model->getKey();

    return get_class($model) . '_' . match (true) {
        $key instanceof \BackedEnum => $key->value,
        $key instanceof \UnitEnum => $key->name,
        default => $key,
    };
}
```

That covers both backed and pure enums and leaves every other key type behaving as it does now.

willpower232: I can appreciate your plight, whilst it is definitely a little niche. Perhaps a PR where the trait gets a `getKeyForAudit()` function which by default returns `getKey()` but then you could override it would solve it and keep things flexible? Then every mention of getKey becomes getKeyForAudit in the package.

kristapscodes: Sounds good, getKeyForAudit() makes sense to me.

One thought — maybe the default implementation of getKeyForAudit() in the trait could handle enums, rather than just returning getKey() as-is? Otherwise I think it could still break by default for enum keys, and you'd probably only notice when it crashes. Something like:

```php
public function getKeyForAudit()
{
    $key = $this->getKey();

    if ($key instanceof \BackedEnum) return $key->value;
    if ($key instanceof \UnitEnum) return $key->name;

    return $key;
}
```

Also, while grepping for getKey() in the package I noticed transitionTo() in Auditable.php does if ($this->getKey() !== $audit->auditable_id). I don't use that method so I can't really judge it, but with an enum key it looks like that comparison might not match, since auditable_id seems to come back from the DB as a string. Might be worth a look while you're in there.

Sorry, I can't take the PR on right now — too much on. On our side we've just overridden getKey() on the two affected models for now.

erikn69: - Hi, it didn't cross my mind that the keys might have casts;
- try #1062.

To add `getKeyForAudit`, perhaps the `Auditable` contract should be modified? Otherwise, how would you ensure the method exists?

kristapscodes: Thanks — #1062 looks like it would work for our case, and nice that getRawOriginal() covers any cast type rather than just enums.

On the contract question — I don't think you'd need to touch it. The observer already does the same for isAuditingEnabled(), which isn't in the Auditable contract either:

`if (method_exists($modelClass, 'isAuditingEnabled') && ! $modelClass::isAuditingEnabled()) {`

So a method_exists() fallback would keep it optional, and the method could move into the contract in a next major if you wanted it guaranteed.

That said, if #1062 lands I don't think getKeyForAudit would be needed at all — it was only a suggestion for making the default safe.

PR #1056, opened 2026-06-16
Refactor audit creation to use savepoint

erikn69:
Wraps the audits creation within a save point transaction using the withSavePointIfNeeded method

Postgres just needs it
For MySQL/SQLServer/SQLite, this is not necessary, but it shouldn't affect anything either.

- Closes #1043


willpower232 (approved):
nice

PR #1059, opened 2026-06-19
Fix Async logging 

### LINKED ISSUE #1058: Async logging does not work
### PHP Version

^8.3

### Laravel Version

13.x

### Package Version

14.0.5

### Description

Hello!
After updating the package earlier this morning, the audit stopped working

**Context:**
- Logging configured to run asynchronously using queue
- Data changes from job & schedules are ignored

**Problem:**
- Audit job dispatched, but no data is recorded unless the `console` config in `config/audit.php` file is true

**What I found:**
I feel like the change in [ProcessDispatchAudit.php](https://github.com/owen-it/laravel-auditing/blob/4fb70d295a04a63ff2c00947a76c42aac1654741/src/Listeners/ProcessDispatchAudit.php#L30) at [4fb70d2](https://github.com/owen-it/laravel-auditing/commit/4fb70d295a04a63ff2c00947a76c42aac1654741) was pretty questionable? 
Image
It calls the static method [Auditable::isAuditingEnabled()](https://github.com/owen-it/laravel-auditing/blob/4fb70d295a04a63ff2c00947a76c42aac1654741/src/Auditable.php#L552) that also checks whether the `audit.console` is true whenever the app is running via queue. Meanwhile the ProcessDispatchAudit is processed using queue in asynchronous configuration.

### Steps To Reproduce

1. Clone this [repo](https://github.com/LevineLighto/reproduce-laravel-auditing-26-06-19)
2. Install & configure laravel env
    ```bash
    composer install
    cp .env.example .env
    php artisan key:generate
    // Fill database information on .env before continuing
    php artisan migrate
    ```
3. Start the app
    ```
    php artisan serve
    ```
4. Start a queue worker
    ```bash
    php artisan queue:work
    -- or
    php artisan queue:listen
    ```
5. Access the app
    ```url
    http://localhost:8000
    ```
6. Input user name and press enter
7. The user panel will refresh with a new entry, but audit panel remains empty
8. Go to config/audit.php, change `console` value to `true`
    (please restart your queue worker if you used `php artisan queue:work`)
9. Go back to your browser
10. Do step 6
11. The user panel will refresh with a new entry, and audit panel should have a new entry (you may need to refresh the audit panel to see the new entry, the audit panel automatically refreshes every 5 seconds)

### Possible Solutions

Change the handle method in [ProcessDispatchAudit.php](https://github.com/owen-it/laravel-auditing/blob/4fb70d295a04a63ff2c00947a76c42aac1654741/src/Listeners/ProcessDispatchAudit.php) to something like this?
```php
public function handle(DispatchAudit $event): void
{
    if (!Config::get('audit.enabled', true)) {
        return;
    }
    Auditor::execute($event->model);
}
```

Anyway, thank you for your great work!

erikn69:
- Closes #1058
- Bug from #1040

Remove check for auditing enabled before executing Auditor.
If the event has already been dispatched, it means that all audit requirements were met;
there is no need to check again.

Same as RecordCustomAudit :
https://github.com/owen-it/laravel-auditing/blob/4fb70d295a04a63ff2c00947a76c42aac1654741/src/Listeners/RecordCustomAudit.php#L8-L14

It works in sync, but we don't take into account async queues.

@willpower232 hi, any comments?

### DIFF
--- a/src/Listeners/ProcessDispatchAudit.php
+++ b/src/Listeners/ProcessDispatchAudit.php
@@ -26,10 +26,6 @@ class ProcessDispatchAudit implements ShouldQueue
 
     public function handle(DispatchAudit $event): void
     {
-        $modelClass = $event->model::class;
-        if (method_exists($modelClass, 'isAuditingEnabled') && ! $modelClass::isAuditingEnabled()) {
-            return;
-        }
         Auditor::execute($event->model);
     }
 }


willpower232 (approved):
I'm far away from a computer for a bit but makes sense from what I can see on my phone 😅

